PR #9, opened 2019-03-12
fix false positive AvoidConcatInAppendTest and new public class AvoidDecimalAndChoiceFormatAsField

stokpop:


stokpop:
i saw that a merge was needed after creating the pull request, let me try to clean that up first

PR #72, opened 2020-08-31
added AvoidJAXBUtil rule, doc updates

stokpop:


jborgers (changes requested):
* In bad example please use '// bad' in stead of '// violation' 
* Can you provide a good example? How to do it right?
* Can you use typeIs() for exact type checking?

jborgers (approved):

PR #139, opened 2021-05-28
fix for issue-125: there are two names for same timeout

stokpop:
there are two names for same timeout: socketTimeout (for HttpComponentsClientHttpRequestFactory) or readTimeout (for RequestConfig)


jborgers (approved):
OK

PR #294, opened 2024-03-01
added AvoidSimpleDateFormat for Kotlin

stokpop:
Because of missing typeIs, the check is a little less strict and the testcase for Unknown class had to be separated from the XmlMapper and ObjectMapper tests. 

Less strict: if there is an import of `fasterxml.jackson` packages and `setDateFormat()` is called, it is allowed.

### DIFF
--- a/src/main/resources/category/kotlin/common.xml
+++ b/src/main/resources/category/kotlin/common.xml
@@ -133,4 +133,47 @@ class Foo {
         </example>
     </rule>
 
+    <rule name="AvoidSimpleDateFormat"
+          class="net.sourceforge.pmd.lang.rule.XPathRule"
+          dfa="false"
+          language="kotlin"
+          message="SimpleDateFormat is used. Since it is thread-unsafe, it needs expensive recreation."
+          typeResolution="true"
+          externalInfoUrl="${doc_root}/JavaCodePerformance.md#idtf01">
+        <description>Problem: java.util.SimpleDateFormat is thread-unsafe. The usual solution is to create a new one when needed in a method. Creating SimpleDateFormat is relatively expensive. &#13;
+            Solution: Use java.time.DateTimeFormatter. These classes are immutable, thus thread-safe and can be made static.
+            (jpinpoint-rules)</description>
+        <priority>2</priority>
+        <properties>
+            <property name="tag" value="jpinpoint-rule" type="String" description="for-sonar"/>
+            <property name="version" value="3.1"/>
+            <property name="xpath">
+                <value><![CDATA[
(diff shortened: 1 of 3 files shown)


src/main/resources/category/kotlin/common.xml
```diff
@@ -133,4 +133,47 @@ class Foo {
...
+        <properties>
+            <property name="tag" value="jpinpoint-rule" type="String" description="for-sonar"/>
+            <property name="version" value="3.1"/>
+            <property name="xpath">
+                <value><![CDATA[
+(: check if java.text imports exists as filter for same named classes in other packages :)
+//ImportHeader[.//T-Identifier[@Text='java'] and .//T-Identifier[@Text='text']][1]/../..
+
+(: if used in setDateFormat function, allow if jackson ObjectMapper or XmlMapper is used :)
+//SimpleIdentifier//T-Identifier[(@Text="SimpleDateFormat") and ancestor::ClassMemberDeclaration
```
jborgers, line 156: waar is deze 'and ancestor::ClassMemberDeclaration' voor nodig?

stokpop: moved it to the main XPath expression

src/main/resources/category/kotlin/common.xml
```diff
@@ -153,7 +153,7 @@ class Foo {
 //ImportHeader[.//T-Identifier[@Text='java'] and .//T-Identifier[@Text='text']][1]/../..
 
 (: if used in setDateFormat function, allow if jackson ObjectMapper or XmlMapper is used :)
-//SimpleIdentifier//T-Identifier[(@Text="SimpleDateFormat") and ancestor::ClassMemberDeclaration
+//ClassMemberDeclaration//SimpleIdentifier//T-Identifier[(@Text="SimpleDateFormat")
```
jborgers, line 156: Je kunt dan toch ook zonder de ancestor af? 
//ClassMemberDeclaration[.//SimpleIdentifier//T-Identifier[(@Text="SimpleDateFormat")
and not (.//T-Identifier[@Text="setDateFormat"] and ...) ]

jborgers (changes requested):

src/main/resources/category/kotlin/common.xml
stokpop: Not sure if I understand. If I start with 
```//ClassMemberDeclaration[.//SimpleIdentifier//T-Identifier[(@Text="SimpleDateFormat")]]```
it matches `private` instead of `SimpleDateFormat`. How to go from there?

PR #301, opened 2024-03-25
Kotlin AvoidMultipleConcatStatements

jborgers:
Avoid multiple statements concatenating to the same property var.
In Kotlin with type inference it is hard to determine the type, String in this case. 
I now support for property var:
-string literal
-explicit type String
-assigned a parameter of type String
-assigned a function of explicit type String (with known issue)
-Note "$value" is supported as it is concatenated like a normal String.

### DIFF
--- a/src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidMultipleConcatStatements.xml
+++ b/src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidMultipleConcatStatements.xml
@@ -5,8 +5,8 @@
         xsi:schemaLocation="http://pmd.sourceforge.net/rule-tests http://pmd.sourceforge.net/rule-tests_1_0_0.xsd">
     <test-code>
         <description>AvoidMultipleConcatStatements</description>
-        <expected-problems>5</expected-problems>
-        <expected-linenumbers>8,22,29,63,70</expected-linenumbers>
+        <expected-problems>12</expected-problems>
+        <expected-linenumbers>8,22,36,52,70,85,103,116,129,143,157,171</expected-linenumbers>
         <code><![CDATA[
 import java.util.Arrays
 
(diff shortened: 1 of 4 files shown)


src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidMultipleConcatStatements.xml
```diff
@@ -0,0 +1,206 @@
...
+        <description>AvoidMultipleConcatStatements</description>
+        <expected-problems>12</expected-problems>
+        <expected-linenumbers>8,22,36,52,70,85,103,116,129,143,157,171</expected-linenumbers>
+        <code><![CDATA[
+import java.util.Arrays
+
+class AvoidMultipleConcatStatementsKt {
+    fun testMultipleConcatDefect() {
+        var logStatement = ""
+        val values = Arrays.asList(*arrayOf("tic", "tac", "toe"))
```
stokpop, line 16: this * spread operator I had to lookup :-) 

in Kotlin speak: 
```
        val values = listOf("tic", "tac", "toe")
```

stokpop: or is it there for a reason to see of that also parses?

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidMultipleConcatStatements.xml
```diff
@@ -0,0 +1,206 @@
...
+
+    fun testMultipleConcatDefect10() {
+        var logStatement = someString();
+        val values = listOf("tic", "tac", "toe")
+        logStatement += values[0]
+        logStatement += values[1] // bad 142
+    }
+
+    fun testMultipleConcatCorrect10() {
+        var logStatement = someInt("");
```
stokpop, line 157: why a test case like this? what is intention of `someInt()` vs `someString()`?

stokpop: Ah, probably to lookup the return value of the method (in same class only)

stokpop (approved):
looks good, maybe also contains candidate for function "classHasMethodWithReturnType"

PS there is another rule in this PR, is that caused by not starting branch from main?

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidMultipleConcatStatements.xml
jborgers: exactly

PR #304, opened 2024-04-05
Add Kotlin AvoidRecreatingDateTimeFormatter

stokpop:
Works on basic unit tests.

Should we create more unit tests to avoid false positives and add more Kotlin syntaxes?


src/main/resources/category/kotlin/common.xml
```diff
@@ -292,5 +291,39 @@ class Foo {
...
+            formatter from a pattern only once, to initialize a constant or instance field.
+            (jpinpoint-rules)</description>
+        <priority>2</priority>
+        <properties>
+            <property name="tag" value="jpinpoint-rule" type="String" description="for-sonar"/>
+            <property name="xpath">
+                <value>
+                    <![CDATA[
+(: check if java.text imports exists as filter for same named classes in other packages :)
+//ImportHeader[.//T-Identifier[@Text='org'] and .//T-Identifier[@Text='joda'] and .//T-Identifier[@Text='time'] and .//T-Identifier[@Text='format']][1]/ancestor::KotlinFile
```
jborgers, line 310: is this about java.text (1st line) or joda.time (2nd line)?
Should it not hold for both cases? So match on at least one of them?
Include threeten.bp.format? Maybe not, old.

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidRecreatingDateTimeFormatter.xml
```diff
@@ -0,0 +1,128 @@
...
+        val formatter2 = DateTimeFormat.forPattern(DEFAULT_DATETIME_FORMAT) // bad
+        val formatter3 = DateTimeFormat.forPattern(DEFAULT_DATETIME_FORMAT).withZoneUTC() // bad
+        val formatter4 = DateTimeFormat.forPattern(DEFAULT_DATETIME_FORMAT).withZone(ZoneId.of("UTC")) // bad
+        val formatter5 =
+            DateTimeFormat.forPattern(DEFAULT_DATETIME_FORMAT).withZone(myTimeZone) // good: has variable parameter
+        return formatter1.print(milliseconds)
+    }
+}
+     ]]></code>
+    </test-code>
```
jborgers, line 127: Can you add a separate test-code with just java.time import? (no joda.time) 
and one without joda.time and without java.time and a custom DateTimeFormatter so without violation?

src/main/resources/category/kotlin/common.xml
stokpop: In the test file there is: 

    val ftor: DateTimeFormatter = org.threeten.bp.format.DateTimeFormatter.ofPattern("yyMMdd") // bad

This will match because we do not check the fully qualified name with packages names: `org.threeten.bp.format.`. So any fully specified custom DateTimeFormatter will have a hit.

We might be able to add this to a function that checks imports and fully classified class names.

For now `threeten` is not explicitly included.

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidRecreatingDateTimeFormatter.xml
stokpop: These are already tests with only java.time and only joda.time.

Added custom DateTimeFormatter no violation test. And also one with just java.time.

src/main/resources/category/kotlin/common.xml
```diff
@@ -366,5 +365,59 @@ class Foo {
...
+//ImportHeader[
+    .//T-Identifier[@Text='java'] and .//T-Identifier[@Text='time'] and .//T-Identifier[@Text='format']
+  ][1]/ancestor::KotlinFile
+
+/TopLevelObject//Expression//T-Identifier[
+    @Text='DateTimeFormatter'
+     (: check if variable function params in init constructor with two params :)
+     and ancestor::PostfixUnaryExpression/PostfixUnarySuffix//ValueArguments[
+           count(ValueArgument) = 2
+           and ./ValueArgument[1]//T-Identifier[@Text = ancestor::FunctionDeclaration/FunctionValueParameters/FunctionValueParameter/Parameter/SimpleIdentifier/T-Identifier/@Text]
```
stokpop, line 411: @jborgers can this part be made more efficient? need to check if both arguments of method are function parameters: is there a xpath expression for "for all ValueArguments there is a function parameter"?

jborgers: Ja, ik denk count(./ValueArgument//T-Identifier[@Text = ancestor::...) = 2

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidRecreatingDateTimeFormatter.xml
jborgers: expected one unit test to fail without the fix

jborgers (changes requested):

src/main/resources/category/kotlin/common.xml
stokpop: thanks, that works!

stokpop (commented):
fixed count(..) = 2 and checked for failing unit tests

jborgers (approved):

jborgers:
still needs to sync from master to pmd7

PR #305, opened 2024-04-12
Added AvoidWideScopeXPathExpression (Kotlin)

stokpop:
First addition. 

Question: if `XPathAPI.eval(doc, "//...")` is used: is also bad? The test case is now only `XPathAPI.eval(doc, "/...")` is _not_ bad.


jborgers (approved):

stokpop:
Merged via kotlin/AvoidWideScopeXPathExpression2 due to some rebase issue in this branch

PR #315, opened 2024-06-18
New rule Kotlin: AvoidImplicitRecompilingRegex

stokpop:
This is a big one. Not all tests and examples work the same for Kotlin compared to Java.

There is no difference for short or long Regexp's as far as I can see.

Basically you should not call `.toRegex()` for each invocation.
Some String methods, like split() have versions without Regex, with String, these are advertised as more effictive.

We could add checks to see if strings are really regexps or could be used as plain string.

Need some help with this one (now not checked in unit test):

```
        val fs = FileSystems.getDefault()
        p = fs.getPathMatcher(PAT_STRING) // bad: violation
```
Where match should be made on the `fs` and the corresponding var to check if it is made of `FileSystems`.

The unit test can be cleaned up and rearranged when all are working.


stokpop:
> > There is no difference for short or long Regexp's as far as I can see.
> 
> The number of characters to process in the compile is different and so the amount of work is, right? Every time a compile if 2 chars is not so bad, of 20 chars is bad. This holds for Kotlin too, I assume.

On the other hand, the Kotlin Regex can easily be used from a (companion object) field.

stokpop:
I think we need to add rules for the default constructor `Regex("bla.*")` and also `toRegex(options)`

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidImplicitlyRecompilingRegex.xml
```diff
@@ -0,0 +1,361 @@
...
+
+    // line 29
+    fun violatingPathMatcher4x(action: String?) {
+        var p = FileSystems.getDefault()
+            .getPathMatcher("(?=\\p{Lu})") // bad: violation, should not be inside method, make static final
+        FileSystems.getDefault().getPathMatcher(PAT_STRING) // bad: violation
+        FileSystems.getDefault().getPathMatcher(dynPatString).matches(path) // bad (2x):  field should be PathMatcher
+
+        val fs = FileSystems.getDefault()
+        p = fs.getPathMatcher(PAT_STRING) // bad: violation
```
jborgers, line 140: To match this line, the last part of xpath expression can be:

```
(: check if java.nio.file.FileSystems or java.nio.file.Path imports exists as filter for same named classes in other packages :)
(: TODO is this a bug: having to use T-FILE for 'file' :)
//ImportHeader[
    (.//T-Identifier[@Text='java']
        and .//T-Identifier[@Text='nio'] and .//T-FILE[@Text='file'] and .//T-Identifier[@Text='FileSystems' or @Text='Path'] or .//T-MULT)
  ][1]/ancestor::KotlinFile
//FunctionDeclaration//PostfixUnarySuffix//T-Identifier[(@Text='getPathMatcher' or @Text='matches')
    and (ancestor::PostfixUnaryExpression/PrimaryExpression/SimpleIdentifier/T-Identifier[@Text='FileSystems'])
    (: not if first param is function parameter :)
    and not(ancestor::PrefixUnaryExpression//CallSuffix/ValueArguments/ValueArgument//SimpleIdentifier/T-Identifier[
       @Text = ancestor::FunctionDeclaration//Parameter/SimpleIdentifier/T-Identifier/@Text])
    (: not param in String template in function param :)
    and not(ancestor::PrefixUnaryExpression//CallSuffix/ValueArguments/ValueArgument//LineStringContent/T-LineStrRef[
       @Text = ancestor::FunctionDeclaration//Parameter/SimpleIdentifier/T-Identifier/concat('$', @Text)])
    or (ancestor::PostfixUnaryExpression/PrimaryExpression//T-Identifier[@Text=
        ancestor::Statement/preceding-sibling::Statement//VariableDeclaration[../Expression//T-Identifier[@Text='FileSystems']]//T-Identifier/@Text])
]
```

The last 3 full lines are new.
Only thing is that it matches not just on 'getPathMatcher' but also on 'PAT_STRING' which I don't understand.

jborgers (changes requested):

src/test/resources/com/jpinpoint/perf/lang/kotlin/ruleset/common/xml/AvoidImplicitlyRecompilingRegex.xml
stokpop: fixed this, needed to add: `[ancestor::NavigationSuffix]` in selector to avoid selecting `PAT_STRING`

```
(: added [ancestor::NavigationSuffix] to avoid PAT_STRING param from being selected too :)
//FunctionDeclaration//PostfixUnarySuffix//T-Identifier[ancestor::NavigationSuffix][@Text='getPathMatcher'
    and (ancestor::PostfixUnaryExpression/PrimaryExpression//T-Identifier[@Text=
              ancestor::Statement/preceding-sibling::Statement//VariableDeclaration[../Expression//T-Identifier[@Text='FileSystems']]//T-Identifier/@Text]
         or ancestor::PostfixUnaryExpression/PrimaryExpression/SimpleIdentifier/T-Identifier[@Text='FileSystems']
 ```

**Comment by stokpop** [comment]
> I think we need to add rules for the default constructor `Regex("bla.*")` and also `toRegex(options)`

also added this

Note that Regex(...) is not so Implicit... can also create a separate rule, but I think it fits with this one

**Comment by stokpop** [comment]
This PR also includes: 
* upgrade to PMD 7.13.0
* make unit tests work for both Java and Kotlin
* remove the maven profiles, both Java and Kotlin can be run from same branch now
* note: for Kotlin merge use shell command: `./merge kotlin`

File: rulesets/kotlin/jpinpoint-kotlin-rules.xml [hunk 2]
```diff
@@ -62,8 +61,118 @@ class Foo {
         </example>
     </rule>
 
+    <rule name="AvoidImplicitlyRecompilingRegex"
+          class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
+          dfa="false"
```
jborgers: dfa to be removed, I changed it in the migration to 7

rulesets/kotlin/jpinpoint-kotlin-rules.xml
```diff
@@ -62,8 +61,118 @@ class Foo {
...
+          language="kotlin"
+          message="String regex method, Pattern.matches or FileSystem.getPathMatcher is used. Implicitly compiles a regex pattern, can be expensive."
+          typeResolution="true"
```
jborgers: typeResolution can also be removed

jborgers (changes requested):
An impressive big one indeed. Just some small changes requested.

rulesets/kotlin/jpinpoint-kotlin-rules.xml
stokpop: dfa is Data Flow Analysis: is this to be removed for all entries, also for Java?

rulesets/kotlin/jpinpoint-kotlin-rules.xml
stokpop: deprecated/hidden: https://pmd.github.io/pmd/pmd_release_notes_pmd7.html#6180

rulesets/kotlin/jpinpoint-kotlin-rules.xml
stokpop: deprecated: https://pmd.github.io/pmd/pmd_release_notes_pmd7.html#6270

jborgers: yes, also for java

stokpop: removed for all

rulesets/kotlin/jpinpoint-kotlin-rules.xml
stokpop: removed for all

jborgers (approved):

PR #321, opened 2024-07-02
318 merge tag properties to comma separated tags property 

stokpop:
On merge it collects all property "tag" in property "tags" with comma separated list.

Seems more convenient to use the separate tags in the original files, and then bring them together on merge. Agree?

--> the original files can be converted as well, right? Seems a bit confusing to have two formats. I'd prefer to have the original file be converted as well.

--> can be done indeed, have xsl scripts counting/reporting the number of tags that were a bit simpler with separate property tags :-)

### DIFF
--- a/rulesets/java/jpinpoint-rules.xml
+++ b/rulesets/java/jpinpoint-rules.xml
@@ -22,10 +22,6 @@
         (jpinpoint-rules)</description>
         <priority>1</priority>
         <properties>
-            <property name="tag" value="jpinpoint-rule" type="String" description="for-sonar"/>
-            <property name="tag" value="performance" type="String" description="for-sonar"/>
-            <property name="tag" value="sustainability-low" type="String" description="for-sonar"/>
-            <property name="tag" value="memory" type="String" description="for-sonar"/>
             <property name="xpath">
                 <value><![CDATA[
 //Expression/PrimaryExpression/PrimaryPrefix/Name[@Image='CDI.current']/../../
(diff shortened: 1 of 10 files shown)


rulesets/java/jpinpoint-rules.xml
```diff
@@ -22,9 +22,6 @@
         (jpinpoint-rules)</description>
         <priority>1</priority>
         <properties>
-            <property name="tag" value="jpinpoint-rule" type="String" description="for-sonar"/>
-            <property name="tag" value="performance" type="String" description="for-sonar"/>
-            <property name="version" value="2.0"/>
```
jborgers, line 27: Why is version disappearing? For PMD 6, version should be in there, for PMD 7 it should be removed.
For generating the jpinpoint-rules-pmd7 file, this is assumed.

jborgers (changes requested):

rulesets/java/jpinpoint-rules.xml
stokpop: Version removal was already done manually in the classification branch, is not done by the merge script. I was not aware that it must be present in PMD6.

stokpop: added back the version for pmd-6

stokpop:
added tags to original files

added the migration java program.. should we keep it or remove it before merge into master?

jborgers (approved):

Issue #309, opened 2024-05-14
Migrate java rules to pmd 7.x.x


emouty:
Because of [breaking changes introduced with pmd 7.x.x](https://docs.pmd-code.org/latest/pmd_userdocs_migrating_to_pmd7.html#im-using-custom-rules), the java ruleset can't be used with this new version.

The main error I've seen is from the package migration of `net.sourceforge.pmd.lang.rule.XPathRule` to `net.sourceforge.pmd.lang.rule.xpath.XPathRule`, but there could be other errors. 

It could be nice to migrate the rulesets to be compatible with this new major version

jborgers:
Hi, thanks for reporting. We are aware of it. This needs quite some time to migrate the rules to 7. We are figuring out how to do this.

jborgers:
The AST provided by PMD7 is incompatible with the one of PMD6. It actually has a lot of structural changes. The xpath expression of each rule needs to be migrated.

jborgers:
Migration is done, except for #476

PR #330, opened 2024-09-09
Migrate java rules to pmd 7.x.x #309: common_std rule number 2: Avoid…

stokpop:
…CalendarDateCreation

The second common_std rule. Note this is a branch that merges into the rule number 1 branch, because it depends on `common_std` being available. So first merge rule 2 into 1 and next merge 1+2 into originating `pmd7-upgrade-7.2.0` branch.

Added a check for `Bar.getInstance().getTime()`: should not be a violation.

I noticed that match on `DateTime` class is by `ClassType/@SimpleName`, so matches both `joda` and `java.time` packages.  Can be made more specific using `typeIs`, but should we do that now?


jborgers:
Hmm, ik heb die andere al gemerged. common_std is er nu bij in base branch, dan moet het met deze toch goed gaan?

stokpop:
Added typeIs and matchesSig. Added unit test lines:

```
                DateTime nowDT3 = new DateTime(Calendar.getInstance().getTime()); // bad
                DateTime nowDT4 = new DateTime(new Date().getTime()); // good
```

The first line now has a "double" hit: see unit test with two times expected line 8. Is that an issue or can we leave that?

jborgers (approved):
Prima zo

PR #333, opened 2024-09-10
Migrate java rules to pmd 7.x.x #309: common_std rule number 5: Limit…

stokpop:
…WildcardStaticImports


stokpop:
conflicts resolved (all pull requests but one in parallel will have resolve issues, as they are all based on the same parent branch)

PR #337, opened 2024-09-12
Migrate java rules to pmd 7.x.x #309: common_std rule number 9: AvoidForEachInStreams

stokpop:
The match on forEach is very generic (`_#forEach(_)`), should also match multiple forEach in streams package: make more explicit or is ok like this?

The lambda check is checking for a log-like statement and if present it skips the check. Should we not add a check that it is the only line in the lambda? Otherwise side effect methods could be present.


jborgers:
Maak jij een Fix Request aan?

jborgers (approved):

stokpop:
added fix request: https://github.com/jborgers/PMD-jPinpoint-rules/issues/341

PR #344, opened 2024-09-18
Pmd7 upgrade 7.2.0 concurrent rules set 01

stokpop:
Migrated first 4 rules from concurrent.

Discovered this syntax: `pmd-java:matchesSig('java.util.Arrays#asList(_*)')` Needed it to match `Arrays.asList("a", "b");`.

Added examples where missing.

Rule `AvoidMutableStaticFields` is most complex in this set. There are still some ommissions and improvements possible. For instance, the checks for collections being wrapped in `Collections.unmodifiableX(...)`: there are more variants of this that might give false positives if not checked. And defining in initializer block `var localVar = new Hashmap()` and wrapping it later in a unmodifiable might not always be detected.


jborgers:
This is a PR to master, not correct. should be to pmd7-upgrade-7.2.0

jborgers:
I didn't know about the (_*) good to know! Can you not use (_,_) I thought that worked.

jborgers:
maybe also merge pmd7-upgrade-7.2.0 into this branch first

stokpop:
both requests done, you can simply edit the PR and switch base branch within github

stokpop:
`(_,_)` did not work in this case as it is a `myMethod(String... strings)` so it can have arbritrary number of params

jborgers (approved):

PR #347, opened 2024-09-23
Pmd7 upgrade 7.2.0 remoting set1

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/remoting/xml/ObjectMapperCreatedForEachMethodCall.xml
```diff
@@ -72,16 +72,16 @@ public class JsonMapperReuse {
 	}
 
 	public static JsonMapper createMapper() {
-	        return JsonMapper.builder()
+	        return JsonMapper.builder() // bad 2x
```
stokpop, line 75: Can double hit be avoided? Does this also create two issues in SonarQube?

stokpop (approved):
Looks good, one question about preventing double hits, but that is not something new. Maybe in PMD7 it is easier to avoid a double hit in places.

PR #365, opened 2024-10-14
Pmd7 upgrade 7.2.0 remoting set3

jborgers:


src/main/resources/category/java/remoting.xml
```diff
@@ -808,7 +808,7 @@ class Good {
             When this HttpHost is then used for a route and stored socketConfig for, port 80 is added for the host and
             the socketConfig is stored with the wrong key and will not be used.
             It is typically difficult to find out if the config is actually used. Note that
-            http://github.com/jborgers/http-client-monitor helps here.&#13;
```
stokpop, line 806: Might also be good to check and point to micrometer and available apache http 4 and 5 metrics (works with spring boot actuator).

src/main/resources/category/java/remoting.xml
```diff
@@ -281,25 +281,17 @@ return HttpClientBuilder.create() // good, setConnectionManager called, pool con
...
+//MethodDeclaration[
+    .//MethodCall/(TypeExpression|VariableAccess)[
+        pmd-java:typeIs('org.apache.http.impl.client.HttpClientBuilder')
+        or pmd-java:typeIs('org.apache.http.impl.client.HttpClients')
+        or pmd-java:typeIs('org.apache.http.impl.nio.client.HttpAsyncClientBuilder')
+    ]
+][
+    .//MethodCall[@MethodName='setConnectionManager']
+]
+(: trying to get to the right line, unfortunately, the chaining seems to make it not work :)
```
stokpop, line 293: hmm, is this always an issue with chaining? should we create an issue in pmd project for this?

stokpop:
Looks, good. We may need to check Apache HttpClient 4 rules if they are compatible with Apache HttpClient 5 (future improvement). Two remarks, but approved.

stokpop (approved):

src/main/resources/category/java/remoting.xml
jborgers: okay, good point, added

src/main/resources/category/java/remoting.xml
jborgers: it pannes in designer as well as in the unit tests, so I guess also in IntellJ. Maybe wait until we checked with the newest pmd (7.6?)

PR #382, opened 2024-11-01
merged all tags from main (pmd6) to pmd7

stokpop:
# common.xml
* `<!-- complex rule TODO ImplementEqualsHashCodeOnValueObjects -->`
* AvoidCalendar is missing in `pmd7`?
* MissingEqualsAndHashCodeWithGetterSetter not in pmd7, also no new tags?

# common_std.xml
* AvoidUnusedAssignments is missing in `pmd7` (todo)

ImplementEqualsHashCodeOnValueObjects - is missing? TODO?

AvoidCalendar, Sonar has a rule to use java.time instead of Calendar etc. ; this rule seems unnecessary, however, not sure, it may be useful in IntelliJ, what do you think?

I added MissingEqualsAndHashCodeWithGetterSetter, correctness, not performance nor sustainability in my view

### DIFF
--- a/src/main-todo/resources/category/java/common_std.xml
+++ b/src/main-todo/resources/category/java/common_std.xml
@@ -15,11 +15,7 @@
         </description>
         <priority>3</priority>
         <properties>
-            <property name="tag" value="jpinpoint-rule" type="String" description="for-sonar"/>
-            <property name="tag" value="unused" type="String" description="for-sonar"/>
-            <property name="tag" value="suspicious" type="String" description="for-sonar"/>
-            <property name="tag" value="replaces-sonar-rule" type="String" description="for-sonar"/>
-            <property name="version" value="2.0"/>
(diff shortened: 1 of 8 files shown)


jborgers:
ImplementEqualsHashCodeOnValueObjects - is missing? TODO?

AvoidCalendar, Sonar has a rule to use java.time instead of Calendar etc. ; this rule seems unnecessary, however, not sure, it may be useful in IntelliJ, what do you think?

I added MissingEqualsAndHashCodeWithGetterSetter, correctness, not performance nor sustainability in my view

jborgers (approved):

stokpop:
Seems both `ImplementEqualsHashCodeOnValueObjects` and `AvoidUnusedAssignments` are complex rules that need to be done still. Not for this PR.

stokpop:
> AvoidCalendar, Sonar has a rule to use java.time instead of Calendar etc. ; this rule seems unnecessary, however, not sure, it may be useful in IntelliJ, what do you think?

Then Sonar Lint plugin should be able to catch that in Intellij, right?

Issue #392, opened 2024-11-05
pmd7 fix request: AvoidImplicitlyRecompilingRegex matches in pmd6, not in pmd7


stokpop:
This one matches in pmd6, not in pmd7:

```java
!userId.matches(FooConstant.USER_REGEX)
```

There is however a check that says "assume remote contant for regex is ok".
Should we turn that around: assume it is not ok? In this case the regex is: 

```java
public static final String USER_REGEX = "^USR\\d+$";
```

PR #393, opened 2024-11-05
pmd7 fix request #392 AvoidImplicitlyRecompilingRegex assume external constants bad?

stokpop:
assume constants for regexps are not simple

For discussion: pmd6 rule had a hit, but that might be a bug as in unit tests other Contant.SEP was assumed good, but should we maybe turn it around? If it is a constant assume it to be not a simple regex? 

```java
import my.app.FooConstant;
class Foo {
      public static String checkUser(String userId){
        if (!userId.matches(FooConstant.USER_REGEX)) { // bad, missing violation or assume good?
            throw new RuntimeException("Invalid userId");
        }
    }
}
```

This was here already, should we turn this into "assumed bad, external"?
```java
public class Try {
    private static final String SMALL = "-.";
    private static final String LARGE = "------.-";
    public static void foo(String query) {
        query.split(Constants.SEP); //assumed good, external
        query.split(SMALL); //good
        query.split(LARGE); //bad
    }
}
```

note: 
```java
public static final String USER_REGEX = "^USR\\d+$";
```

### DIFF
--- a/src/main/resources/category/java/common.xml
+++ b/src/main/resources/category/java/common.xml
@@ -168,12 +168,23 @@ and not (preceding-sibling::SwitchLabel[@Default=true()])]
     or pmd-java:matchesSig('java.lang.String#matches(java.lang.String)')
     or pmd-java:matchesSig('java.nio.file.FileSystem#getPathMatcher(java.lang.String)')
 ]
-/ArgumentList/*[1][(self::StringLiteral and string-length(@Image) > 5 and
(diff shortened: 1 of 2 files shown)


jborgers (changes requested):
For this rule I'd like to keep assuming external regex strings are short to avoid false positives.

stokpop:
As discussed, will not merge this as to avoid false positives.

Issue #396, opened 2024-11-05
Fix request for pmd7 AvoidUserDataInSharedObjects false positive


jborgers:
```
import lombok.Getter;
import lombok.Setter;
import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.stereotype.Component;

@Getter
@EqualsAndHashCode
@Component
@ConfigurationProperties(prefix = "hazelcast")
public class HazelcastProperties implements BaseHazelcastProperties {
    private static final String HAZELCAST = "hazelcast.";
    @Value("${"+ HAZELCAST + "mapNameAccount}")
    private String mapNameAccount; // <--- false positive
}
```

jborgers:
@ConfigurationProperties and no @Setter should be ok

PR #397, opened 2024-11-05
Fix request for pmd7 AvoidUserDataInSharedObjects false positive #396

jborgers:


stokpop (approved):
I also saw some false hits on contants with `MAX`. Hopefully fixed with this :-)

Issue #407, opened 2024-11-12
Fix Request: AvoidForEachInStreams false positive


jborgers:
It also matches on Collection.forEach(). That forEach takes a lambda, however, it is not a stream which can be made parallel. It makes the iteration internal compared to the foreach statement on the collection, it is not to be in a functional expression.

jborgers:
Note this is a false positive in both pmd6 and pmd7 version. Only fixed it in pmd7 branch.

Issue #341, opened 2024-09-16
Fix Request: AvoidForEachInStreams - do not allow log with non-log statements


stokpop:
Log lines are considered good in for each. 

But if there are statements next to logging, it should be "bad".

Example, add the process(l) to this test case in `violation: forEach in streams`:

```java
        letters.stream().forEach(l -> {
            process(l);
            log.info(l);
        });
```

PR #409, opened 2024-11-14
Fix: pmd7 AvoidForEachInStreams false positives #407, and #341 and #410

jborgers:


jborgers:
Covers #407, #341 and #410

stokpop (approved):

PR #414, opened 2024-11-19
pmd7 fix request #413 AvoidUnnecessaryStringBuilderCreation while loop - fix false positive

### LINKED ISSUE #413: Fix Request: pmd7 AvoidUnnecessaryStringBuilderCreation false positive with while loop
```java
    String foo(String value, int len) {
        StringBuilder zeroes = new StringBuilder();

        while (zeroes.length() + value.length() < len) {
            zeroes.append('0');
        }

        return zeroes.append(value).toString(); // <-- false positive, zeroes used in while loop
    }
```

stokpop:
Tried to simplify xpath.

All unit test are green.

Can we think of more while/stream/for loops that we missed?
Watch for more false positives with new xpath?

One unit test matches same line 3 times: `append` is chained 3 times, can we make it that it matches each separate line?


jborgers (approved):

stokpop:
ended up in rebase hell, use PR #495 instead

Issue #430, opened 2024-11-25
Fix Request: false positive in AvoidUnconditionalBuiltLogStrings


jborgers:
Message is logged and returned, so a flase positive:
```
import lombok.extern.slf4j.Slf4j;

@Slf4j
class Foo {
    String bar() {
        StringBuilder messageBuilder = new StringBuilder();
        messageBuilder.append("-"); // <-- false positive
    
        log.info("message: {}" , messageBuilder);

        return messageBuilder.toString();
    }
}
```

PR #431, opened 2024-11-25
Fix: false positive in AvoidUnconditionalBuiltLogStrings #430

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidUnconditionalBuiltLogStrings.xml
```diff
@@ -308,6 +308,26 @@ public class MyProducer {
...
+        <code><![CDATA[
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+class Issue430 {
+    String bar() {
+        StringBuilder messageBuilder = new StringBuilder();
+        messageBuilder.append("-"); // <-- false positive
+
+        log.info("message: {}" , messageBuilder);
```
stokpop, line 327: here in this particular code case: when info level is active, the messageBuilder.toString() is actually called twice: for logging and the return value... might be for another rule, there is some unnecessary overhead. But might be difficult to catch and not worth the effort for a rule  (?)

stokpop (approved):
See comment, not for this rule, but there is some unnecessary overhead. ;-)

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidUnconditionalBuiltLogStrings.xml
jborgers: yeah, good point, maybe worth a rule, but not high priority

Issue #434, opened 2024-11-28
Fix Request: AvoidMultipleConcatStatements spurious violations and missing cases


jborgers:

jborgers:
```
class Issue435 {

    private String[] values = {"tic", "tac", "toe"};

    String bad1() {
        String logStatement = "";
        logStatement += values[0];
        logStatement = logStatement + values[1]; // bad <-- missed case
        return logStatement; 
    }

    String bad2() {
        String logStatement = "";
        logStatement = logStatement + values[1];
        logStatement += values[0]; // bad <-- missed case
        return logStatement; 
    }

    String bad3() {
        String logStatement = "";
        logStatement = logStatement + values[0];
        logStatement += values[1] + "-"; // bad
        logStatement += values[2]; // <-- 'false positive', spurious
        logStatement += "-bla"; // <-- 'false positive', spurious
        return logStatement; // bad
    }

    String bad4() {
        String logStatement = "";
        logStatement = logStatement + values[0];
        logStatement = logStatement + values[1] + "-"; // bad
        logStatement = logStatement + values[2]; // <-- 'false positive', spurious
        logStatement = logStatement + "-bla"; // <-- 'false positive', spurious
        return logStatement; // bad
    }
}

```

PR #436, opened 2024-11-28
Fix: AvoidMultipleConcatStatements spurious violations and missing cases #434

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidMultipleConcatStatements.xml
```diff
@@ -36,7 +36,7 @@ public class AvoidMultipleConcatStatementsTest {
 
 		logStatement = logStatement + values.get(0);
 		logStatement = logStatement + values.get(1); // bad
-        logStatement = logStatement + values.get(2); // bad
+        logStatement = logStatement + values.get(2); // not reported to prevent violation overload
```
stokpop, line 39: so if the case of `get(1)` gets fixed, the second one is reported in new PMD run?

stokpop (approved):
see question

Issue #438, opened 2024-12-02
Fix Request: ImplementEqualsHashCodeOnValueObjects has false positives


jborgers:
In some cases it is unlikely that object are ever used in a collection or called equals on:
* if implements javax.ws.rs.ext.ContextResolver
* if implements javax.ws.rs.client.ClientRequestFilter

PR #443, opened 2024-12-03
Fix: ImplementEqualsHashCodeOnValueObjects has false positives #438

jborgers:


stokpop (approved):
I think we will find other cases to be added to the ignore list in the future :-)

PR #448, opened 2024-12-05
Fix Request: false positive in UseClearKeyGeneratorName #446

### LINKED ISSUE #446: Fix Request: false positive in UseClearKeyGeneratorName
```
import org.springframework.cache.interceptor.KeyGenerator;

public class CustomerCacheKeyGenerator implements KeyGenerator { // <-- false positive
   public Object generate(Object target, Method method, Object... params) {
        // build key and return it
   }
}
```
The class name is not checked as documented for the rule.

jborgers:
to correct branch, pmd7: fixes #446


stokpop (approved):
There can be more generic names like "Generator", leave that for when found in code scans?

yes, good point, agreed

PR #464, opened 2024-12-24
New rule: AvoidUnclearHttpRouteSecurity #364

### LINKED ISSUE #364: Rule Request: AvoidUnclearHttpRouteSecurity
Typically, in our situation we use https, and org.apache.http.conn.routing.HttpRoute and org.apache.hc.client5.http.HttpRoute both default to http. Make the security explicitly by requiring a version of the constructor which has the `secure` boolean. 

If we can determine the https was (highly likely) intended to be used, by use of org.apache.http.conn.ssl.SSLConnectionSocketFactory or  org.apache.hc.core5.http.nio.ssl.BasicClientTlsStrategy, org.apache.hc.core5.ssl.SSLContextBuilder, or javax.net.ssl.SSLContext, 
then this would be a BLOCKER.

So, one rule for expressing it explicitly for clarity and preventing a possible bug, and one for a likely bug.

Probably, this can be a combined rule for version 4 and 5.

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -551,6 +551,37 @@ class Good {
 ```
 **See:** [IUOXAR09: XML related `XXXFactory.newInstance()` is called repeatedly.](JavaCodePerformance.md#IUOXAR09)   
 
+#### IBI24
+**Observation: HttpRoute is constructed without specifying the secure argument.**   
+**Problem:** For Apache HttpRoute, if you don't specify whether the route is secure, the default of non-secure is taken. This is unclear. If you use it to configure a connection manager for that route, and the actual route is secure,
+the key does not match, and you don't configure anything. Then, the default number of connections (2) will be used which may cause requests to wait long for a connection to become available (without a proper timeout) resulting in bad responsiveness.
```
stokpop, line 557: better? "and you don't configure anything" -> "and the intended configuration will not be used"

docs/JavaCodePerformance.md
```diff
@@ -551,6 +551,37 @@ class Good {
...
+
+public class Foo {
+  void bar() {
+    for (Route route : connectionProperties.routes()) {
+      URL url = new URL(route.host());
+      HttpHost httpHost = new HttpHost("https", url.getHost(), getPort(url));
+      HttpRoute httpRouteBad1 = new HttpRoute(httpHost); // bad
+      HttpRoute httpRouteBad2 = new HttpRoute(httpHost, null); // bad
+      boolean secure = true;
+      HttpRoute httpRouteGood = new HttpRoute(httpHost, null, secure);
```
stokpop, line 575: add "// good"?

stokpop (changes requested):
From the issue #364 I thought this was a security issue. But it actually is a configuration issue that impacts performance, I think the problem description makes this clear.

Makes me wonder if in test envs where http might be used instead of https, the config might differ. This is an issue when the test env is a performance test env, which should actually be prod like, so also https.

docs/JavaCodePerformance.md
```diff
@@ -551,6 +551,37 @@ class Good {
...
+import java.net.URL;
+
+public class Foo {
+  void bar() {
+    for (Route route : connectionProperties.routes()) {
+      URL url = new URL(route.host());
+      HttpHost httpHost = new HttpHost("https", url.getHost(), getPort(url));
+      HttpRoute httpRouteBad1 = new HttpRoute(httpHost); // bad
+      HttpRoute httpRouteBad2 = new HttpRoute(httpHost, null); // bad
+      boolean secure = true;
```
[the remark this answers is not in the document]
jborgers: check

docs/JavaCodePerformance.md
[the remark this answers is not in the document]
jborgers: check

jborgers: check

docs/JavaCodePerformance.md
jborgers: check

PR #465, opened 2024-12-28
Fix: UnconditionalOperationOnLogArgument (and 2 other log rules) false positive #455 - exclude private logXX method which are only called conditionally

### LINKED ISSUE #455: Fix Request: UnconditionalOperationOnLogArgument (and other log rules) false positive
This rule doesn't check if it is executed in a method starting with 'log'. A mehod starting with 'log' is likely called within an if (isTraceEnabled) or similar, so ignore those cases.

The rule UnconditionalCreatedLogArguments actually does that check. 

Look into the other log rules, too.

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/UnconditionalCreatedLogArguments.xml
```diff
@@ -226,4 +226,44 @@ public class MyProducer {
...
+   void callerCond(UUID id) {
+        if (log.isDebugEnabled()) {
+            logDelete();
+        }
+    }
+}
+
+@Slf4j
+public class Issue455_AlsoUnCond {
+    private void logDelete(UUID id) {
```
stokpop, line 253: also add test case with non-private log method?

stokpop (approved):
Does the method name need to start with "log" or can it also be any name? (It might make the rule too "heavy" in processing if you allow any name: for all unconditional log cases check if method call is in conditional block.)

> in the cases I have seen in the code, the method starts with 'log', so that seems a good selection to me.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/UnconditionalCreatedLogArguments.xml
jborgers: check

PR #466, opened 2024-12-31
New Rule: AvoidSideEffectsInStreams #461

### LINKED ISSUE #461: Rule Request: AvoidSideEffectsInStreams
Streams are designed for functional programming: return a result of the streaming pipeline, possibly in parallel.
By the spec, everything that does not contribute to the result, may be optimized away. So, side effects are not guaranteed to be executed.
See [stream/package#SideEffects](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/package-summary.html#SideEffects)

The following example modifies method-local state in the peek step to use it in a map step. It is made thread-safe with Atomic, so when making it a parallelStream, it seems to be safe. However, the threads share the one atomicReference, so share the holder of the endpoint thereby mixing the endpoints up,  this would result in a subtle concurrency bug.
```
public List<String> getEndpointsPingInfo(String... pingEndpoints) {
        AtomicReference<String> currentEndpoint = new AtomicReference<>();
        return Arrays.stream(pingEndpoints)
                .peek(endpoint -> currentEndpoint.set(endpoint))
                .map(String::toLowerCase)
                .map(pingInfo -> addEndpointInfo(pingInfo, currentEndpoint.get()))
                .toList();
}
```

jborgers:


docs/JavaCodeQuality.md
```diff
@@ -459,6 +459,40 @@ class AvoidForEachInStreams {
...
+
+@Slf4j
+class AvoidSideEffectsInStreams {
+    final String[] array = new String[]{"a", "b", "c"};
+
+    public List<String> getEndpointsInfo(final List<String> list, String... endpoints) {
+        AtomicReference<String> currentEndpoint = new AtomicReference<>();
+        return Arrays.stream(endpoints)
+                .peek(endpoint -> currentEndpoint.set(endpoint)) // bad
+                .peek(endpoint -> array[0] = endpoint)           // bad
```
stokpop: peek should be used for debugging only, so maybe good to add "bad" examples with map or filter with side effects too... now it seems peek is bad (which it is) e.g. (based on example in the nice Goetz article): 

```java
.map(endpoint -> { array[0] = endpoint; endpoint })
```
-> good point

stokpop (changes requested):
Looks good. Coverage is limited to some simple collections related methods such as `put`. Maybe add some others, see comments.

docs/JavaCodeQuality.md
jborgers: I think that could be a good improvement to the rule, a next version

src/main/resources/category/java/common_std.xml
```diff
@@ -459,4 +459,45 @@ class Foo {
...
+        </description>
+        <priority>3</priority>
+        <properties>
+            <property name="tags" value="bad-practice,data-mix-up,jpinpoint-rule,pitfall" type="String" description="classification"/>
+            <property name="xpath">
+                <value><![CDATA[
+(: matches java.lang.Stream and java.util.stream.IntStream like types :)
+//MethodCall[pmd-java:typeIs('java.util.stream.BaseStream')]
+(: modifying methods :)
+[ArgumentList[.//(MethodCall|MethodReference)[@MethodName=('set','add','put')]/VariableAccess or .//AssignmentExpression/ArrayAccess/*[1][pmd-java:nodeIs('VariableAccess')]
```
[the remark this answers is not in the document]
jborgers: Covered all of those. 

To think about, all access is not appropriate according to Goetz, not just modifying. 
Maybe extend the rule? another rule? Many occurrences?

PR #479, opened 2025-01-21
Fix Request: false positive in MDCPutWithoutRemove - doFinally #326  - put in a called method #473

### LINKED ISSUE #326: Fix Request: false positive in MDCPutWithoutRemove - doFinally
Example:
image

### LINKED ISSUE #473: Fix Request: False positive for MDCPutWithoutRemove - put in a called method
If we add a value to the MDC in the child method and remove it in the finally block of the parent method, It still complains about the issue. For this rule to work, value should be added and removed from MDC in the same method. Please check

public void A () {
try {
  MDC.put("a"); //Does not complain
  B();
} finally {
  MDC.remove("a");
  MDC.remove("b");
 }
}

private void B() {
  MDC.put("b"). //complain with Sonar issue
}

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/MDCPutWithoutRemove.xml
```diff
@@ -156,4 +156,114 @@ public class TracingInterceptor extends HandlerInterceptorAdapter {
...
+import org.slf4j.MDC;
+import org.springframework.security.web.server.authentication.ServerAuthenticationConverter;
+import reactor.core.publisher.Mono;
+import org.springframework.web.server.ServerWebExchange;
+
+class Issue326 {
+        private static final String MDC_REQUEST_ID_KEY = "Request-id";
+        private final ServerAuthenticationConverter authenticationConverter;
+
+       public Mono<Void> filter(ServerWebExchange serverWebExchange) {
```
stokpop, line 172: I have some questions: 
- is doFinally always called, like `finally` in Java?
- is doFinally lambda executed on some thread, or is the MDC available on the doFinally thread? The MDC should be copied and also be cleared in proper way? MDC and reactor are no friends...

Googling for this shows similar questions and (it seems) no simple answers...

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/MDCPutWithoutRemove.xml
```diff
@@ -156,4 +156,114 @@ public class TracingInterceptor extends HandlerInterceptorAdapter {
...
+        try {
+            MDC.put("a-key", "a-value"); // ok
+            bar();
+        } finally {
+            MDC.clear();
+        }
+    }
+
+    public void bar() {
+        MDC.put("b-key", "b-value"); // bad, method public
```
stokpop, line 263: should this also be the `a-key`? even if bar() is private it still fails because the key is different?

stokpop (changes requested):

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/MDCPutWithoutRemove.xml
stokpop: Intention of given code is clear, but wonder if it works/is best practice...

jborgers: Yeah, good question. I assumed the requestor is correct in how he uses it.
I see in javadoc: doFinally does not operate by default on a particular [Scheduler](https://reactivex.io/RxJava/2.x/javadoc/2.0.8/io/reactivex/Scheduler.html).

jborgers: I am asking the requestor

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/MDCPutWithoutRemove.xml
jborgers: Not sure what you mean?

stokpop: Oh, never mind, I see the finally is doing `MDC.clear()` so all keys are removed.

If there was only a `MDC.remove("a-key")` instead, I assume the `MDC.put("b-key", "b-value")` is still bad if turned to method private because the b-key is not removed: I think this is already covered in Issue473_d.

jborgers: right :-)

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/MDCPutWithoutRemove.xml
jborgers: Zeljko is not sure on second thoughts. He will do some tests to find out.

PR #481, opened 2025-01-23
New Rule: AvoidInfiniteRecursion #478

### LINKED ISSUE #478: Rule Request: AvoidInfiniteRecursion
Recursion without a proper stop condition is risky. It can lead to infinite looping leading to stack overflow or out of memory crashes. 
Bad example: 
```
   private Mono<Void> waitForLock() {
        return Mono.defer(() -> {
            if (lock.compareAndSet(false, true)) {
                return Mono.empty();
            } else {
                return Mono.delay(Duration.ofMillis(10))
                    .then(waitForLock());
            }
        });
    }
```
This code lead to an OutOfMemoryError when a call to a backend using this waitForLock took very long, because of unavailability. The solution is to have a stop condition, wait for some time but not infinitely.
Good example:
```
   private Mono<Void> waitForLock(int attempt) {
        return Mono.defer(() -> {
            if(attempt >= 20) {
                return Mono.error(new JWKException("Unable to acquire JWK refresh lock"));
            } else if (lock.compareAndSet(false, true)) {
                return Mono.empty();
            } else {
                return Mono.delay(Duration.ofMillis(50))
                    .then(waitForLock(attempt+1));
            }
        });
    }
```

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -2511,12 +2511,62 @@ class SingletonBMGood {
...
+  private void foo() {
+    boolean success = tryRemoteCall();
+    if (!success && attemptsLeft.decrementAndGet() >= 0 ) {
+      delay(10, MILLISECONDS);
+      foo();
+    }
+  }
+}
+```
+**See:** [Java recursion by Baeldung](https://www.baeldung.com/java-recursion)
```
stokpop: do we have a link with less commercials? :-) Maybe this one: https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/examples-Java-recursion-recursive-methods

This also advices to not use recursion in production code. Can this retry example also be done without recursion? That would be better solution. Max attempts with retries is a good idea and also supported by @Retry annotations. But max calls in a calculation like Fibonacci algos? What number would you use?

Maybe better advice to avoid recursion if possible (until Java solves this or you use alternatives that are not really recursive in call stack nature: https://stackoverflow.com/questions/860550/stack-overflows-from-deep-recursion-in-java). So "AvoidRecursion"?

stokpop: Note that the rule can also be applied to loops: is there in all cases an escape from the loop? (Can you make a rule for that?)

stokpop (changes requested):
Have some doubt, should it not be: AvoidRecursion? 

JB: Well, agreed in general recursion is not such a good idea, for performance especially, but it only leads indisputable to real bad stuff like out of memory in case it is infinite.

docs/JavaCodePerformance.md
jborgers: I'don't accept unnecessary cookies, I guess that is the reason I only have a few commercials. I can add that link. 

Max attempts for algos like Fibonacci should be a large enough number to accommodate the calculations it is used for, and it should be tested.

Good point. A rule to AvoidRecursion may be a good idea, then on a lower severity level.  
Good point. A rule for a loop would also be useful.

jborgers: I will add to the description that recursion may not be a good idea in general.

PR #482, opened 2025-01-23
Pmd7 Fix request: false postive UnconditionalOperationOnLogArgument #468

### LINKED ISSUE #468: Fix request: false postive UnconditionalOperationOnLogArgument
False positive when there are more conditions in if statement:

```java
    public void testGood(String faultCode) {
        if (ids.contains("some-id") && log.isDebugEnabled()) {
            log.debug("Getting [{}]", faultCode.toUpperCase(Locale.US));  // good, but false positive
        }
    }
```

stokpop:


jborgers (approved):
what about other operations like xor, '&' and '|'? You want to deal with them? or just assume they are not used? Maybe another rule checks?

stokpop:
> what about other operations like xor, '&' and '|'? You want to deal with them? or just assume they are not used? Maybe another rule checks?

As these are all bitwise operators, I would not expect these in if statements the contain an boolean isLoggingEnabled() call: there might already be rules that warn/inform about this?

Anyway: to avoid changing it from '||' to '|' (bitwise or), excluded '|' and '^' as well.

jborgers (approved):

PR #485, opened 2025-01-27
Fix : false positives for AvoidInfiniteRecursion #484

### LINKED ISSUE #484: Fix request: false positives for AvoidInfiniteRecursion
These contructs give false positives:

```java
@Service
@RequiredArgsConstructor
public class RepoClient {

    private final DetailRepository detailRepository;

    public EventDocument save(EventDocument documentToSave) {
        return detailRepository.save(documentToSave);
    }
}
```

and this one:

```java
@Builder
@Getter
@EqualsAndHashCode
@ToString
public final class IssueRecursion implements Comparable<IssueRecursion>{

  private final String name;

  @Override
  public int compareTo(IssueRecursion issue) {
    return name.compareTo(issue.getName());
  }
}
```

and this one:

```java
public class Listener extends ListenerSupport {

    @Override
    public boolean open(Context context) {
        return super.open(context);
    }
}
```

and this one: 

```java
public class Deserializer {

    public Map deserialize(JsonParser p, Context c) {
        ....
        return target;
    }

    public Map deserialize(JsonParser p) {
        Context c = new Context();
        return this.deserialize(p, c);
    }
}
```

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
```diff
@@ -249,4 +249,86 @@ class IterationsBetter {
      ]]></code>
     </test-code>
 
+    <test-code>
+        <description>Issue484 No violation: false positives on same method name but other class</description>
```
stokpop, line 253: Did you also have a test for overloaded methods, as this example (I had added it to the issue a bit later, did you see it?) 

```java
public class Deserializer {

    public Map deserialize(JsonParser p, Context c) {
        ....
        return target;
    }

    public Map deserialize(JsonParser p) {
        Context c = new Context();
        return this.deserialize(p, c);
    }
}
```

stokpop (changes requested):

src/main/resources/category/java/common.xml
```diff
@@ -2209,7 +2209,7 @@ class ValueGood {
     ((.//UnaryExpression[@Operator=('++', '--')]|.//InfixExpression[@Operator=('+', '-')]/NumericLiteral[@ValueAsInt=1]|.//AssignmentExpression[@Operator=('+=', '-=')]/NumericLiteral[@ValueAsInt=1])
```
stokpop, line 2209: The rule "message" of `AvoidInfiniteRecursion` is empty, this breaks loading the plugin in SonarQube.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
jborgers: No, did not see. A new case, overloaded method of the same class

src/main/resources/category/java/common.xml
jborgers: good point, was forgotten

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
jborgers: I added additional checking whether the number of arguments match. Actually checking if the argument types match is very difficult, I didn't find a reasonable way in XPath. Unfortunately, there doesn't seem to be a way to access Function like in the designer. OTOH calling an overloaded method with the same number of arguments is rather confusing and bad practice IMHO.

Issue #483, opened 2025-01-24
Fix request: false positive for HashCodeOnlyCallsSuper -- not when equals also calls only super


stokpop:
When equals also only calls super.equals no violation.

```java
    @Override
    public boolean equals(Object o) {
        return super.equals(o);
    }

    @Override
    public int hashCode() {
        return super.hashCode();  // <---- false positive
    }
```

Note: if there are object fields not used in equals/hashCode... it might be an issue or not? Other rule? It is not always clear if fields should be logically part of equals/hashCode, it depends.

jborgers:
There is another rule for equals in value objects, some conditions apply, maybe at least 2 fields (not sure) and corresponding getter(s). Every non-dependent field should take part in the equals.

PR #486, opened 2025-01-27
Fix : false positive for HashCodeOnlyCallsSuper -- not when equals also calls only super #483

### LINKED ISSUE #484: Fix request: false positives for AvoidInfiniteRecursion

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/HashCodeOnlyCallsSuper.xml
```diff
@@ -145,4 +145,22 @@ class EqHashTryout2 {
      ]]></code>
     </test-code>
 
+    <test-code>
+        <description>no violation: Issue false positive for HashCodeOnlyCallsSuper -- not when equals also calls only super #483 </description>
+        <expected-problems>0</expected-problems>
+        <code><![CDATA[
+class Foo {
+    @Override
+    public boolean equals(Object o) {
```
stokpop, line 155: might add a case where equals contains another line, e.g. a valid issue..?

Can also be an enum compare, e.g. `status == o.status && super.equals(o)`?

stokpop (changes requested):
question show that it works when `equals()` contains another line (is it always a `MethodCall`?) than `super.equals(o)`? Can also be an enum compare, e.g. `status == o.status && super.equals(o)`?

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/HashCodeOnlyCallsSuper.xml
jborgers: Good point

stokpop (approved):

PR #491, opened 2025-01-28
Fix: AvoidInfiniteRecursion false positives, call with same number of parameters #490

### LINKED ISSUE #490: Fix Request: AvoidInfiniteRecursion false positives, call with same number of parameters
False positive if a parameter is used in the call with calling a method on it:

```
public class Foo {
    public String translate(Request request) {
        return translate(request.getAccounts()); // good
    }
    public String translate(List<String> accounts) {
        return convert(accounts);
    }
}
```
also with an extra call in between:
```
public class Foo2 {
    public String translate(Request request) {
        return translate(Optional.ofNullable(request.getAccounts())); // good
    }
    public String translate(List<String> accounts) {
        return convert(accounts);
    }
}
```

And if any parameter is used in the call in another position:
```
public class Foo7 {
    public String translate(Request request, String str) {
        Request req;
        return translate(str, req); // good
    }
    public String translate(String str, Request request) {
        return convert(request);
    }
}
```

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
```diff
@@ -381,6 +381,106 @@ class Issue484_onThisAttributeCall {
...
+    }
+    public String translate(String str, Request request) {
+        return convert(request);
+    }
+}
+
+public class Foo8 {
+    public String translate(Request request, String str) {
+        String str2;
+        Request req;
```
stokpop, line 467: req is unused here? what is difference with Foo9? There req is used but type cannot be determined or so? Can you explain in code with comment?

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
```diff
@@ -381,6 +381,106 @@ class Issue484_onThisAttributeCall {
...
+                     ]]></code>
+    </test-code>
+
+    <test-code>
+        <description>No violations - Issue484-more additional false positives if a parameter is used in the call with calling a method on it, and if any parameter is used in the call in another position</description>
+        <expected-problems>4</expected-problems>
+        <expected-linenumbers>33,43,55,87</expected-linenumbers>
+        <code><![CDATA[
+import java.util.List;
+import my.lib.Request;
```
stokpop, line 394: This is an unresolved type? What if all can be resolved: would you be able to compare types in argument lists of methods? (I do no see that in current XPath: check list of types of method params compared to the method definiition list of parameter?).

stokpop (approved):
Might be good enough for now, have some questions about use of types to match method call.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
jborgers: okay, done. Removed `req`, not used.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidInfiniteRecursion.xml
jborgers: In MethodCall the types are known (with imports), but unretrievable in XPath, irrespective of imports. You would have to look them up by XPath matching on variable name, at definition site. You can do pmd-java:typeIs(literalFullyQualifiedClassName) for the param but no way to get the type of the param to compare.

stokpop: Ah yes, had to try it in the PMD designer to see what you mean. Would be nice if java functions in XPath could return nodes to compare instead of just being a boolean with a static value.

jborgers: yes, indeed. That you can get it like `Node[pmd-java:Type = otherNode/pmd-java:Type]` or just as attribute `Node/@Type`. 
I looked in the pmd own rules xml files, in the  designer source and pmd source but I didn't find a clue that they can be retrieved.
Maybe we should ask in the pmd forum and feature request.

PR #494, opened 2025-01-29
Fix: AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects - static inner class of @Component #492

### LINKED ISSUE #492: Fix Request: AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects - static inner class of @Component
example:

```
@Component
public final class SomeUtil {
    public static class SomeID {
        private String userID;
        public String getUserID() {
            if (userID == null) {
                userID = initialUserID(); // <- false positive
            }
            return userID;
        }
   }
}
```

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects.xml
```diff
@@ -556,4 +556,28 @@ class Foo {
...
+        private String userID;
+
+        public String getUserID() {
+            if (userID == null) {
+                userID = initialUserID(); // <- false positive
+            }
+            return userID;
+        }
+   }
+}
```
stokpop, line 579: what for 

```java
    public String doIt() {
        SomeID someID = new SomeID();
        someID.userID = "someUserID"; // <- false positive??
        return someID.toString();
    }
```

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects.xml
```diff
@@ -556,4 +556,28 @@ class Foo {
...
+    <test-code>
+        <description>Avoid unguarded assignment to non-final fields in shared objects: AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects - static inner class of @Component #492</description>
+        <expected-problems>0</expected-problems>
+        <code><![CDATA[
+import org.springframework.stereotype.Component;
+
+@Component
+public final class Issue492Util {
+
+    public static class SomeID {
```
stokpop, line 569: what if `public class SomeID`, e.g. non-static: does it make it different?

stokpop (changes requested):
2 questions

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects.xml
jborgers: yeah, a false positive. If it occurs in practice, lest solve it then, and consider all concurrency rules like this.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects.xml
jborgers: non-static means the inner class instance is part of the outer class instance, so when outer is a @ Component, singleton, inner instance is part of that, so shared and needs to be thread-safe.

stokpop (approved):

Issue #498, opened 2025-01-30
Fix Request: AvoidInfiniteRecursion false positive for same number of params, other types


stokpop:
Example:

```java
    public void tellNext(TbMsg msg, String relationType) {
        tellNext(msg, Collections.singleton(relationType)); // <-- false positive
    }

    public void tellNext(TbMsg msg, Set<String> relationTypes) {
       ... 
        nodeCtx.tell(new TellNextMsg(msg, relationTypes));
    }
```

Possible solution: when multiple methods with same number of params in same class, assume it is no recursion

More cases with singe/collection, e.g. feature/feature**s**

```java
    public Map<Feature, Value> checkStateUpdated(Set<Feature> features) {
        Map<Feature, Value> result = new HashMap<>();
        for (Feature feature : features) {
            Pair<Feature, Value> tmp = checkStateUpdated(feature);
        }
        return result;
    }

    private Pair<Feature, Value> checkStateUpdated(Feature feature) {
        Value value = createValue(feature);
        return Pair.of(feature, value);
    }

```

PR #504, opened 2025-01-31
Fix: AvoidInfiniteRecursion false positive for same number of params, other types #498

jborgers:


rulesets/java/jpinpoint-rules.xml
```diff
@@ -324,10 +324,12 @@ class Good {
 ])]//MethodCall[@MethodName = ancestor::MethodDeclaration/@Name][not(VariableAccess or FieldAccess or SuperExpression or TypeExpression or MethodCall or AmbiguousName)]
 (: only if the number of parameters match: not an overloaded method with difference in numbers :)
 [ArgumentList/@Size = ancestor::MethodDeclaration/FormalParameters/@Size]
-(: not if a parameter is used in the call with calling a method on it :)
-[not(ArgumentList//MethodCall/VariableAccess/@Name = ancestor::MethodDeclaration/FormalParameters//VariableId/@Name)]
+(: not if a parameter is used in the call with calling a method on it or with it as param :)
```
stokpop, line 327: bit cryptic: "or with it as param" -> what does it mean?

stokpop (approved):
one request for clarification in comment

I expect not all false positives will be gone: let's see how many are left after this

rulesets/java/jpinpoint-rules.xml
jborgers: like getId(param) besides param.getId()

PR #508, opened 2025-02-03
Fix: AvoidInfiniteRecursion - this.field false positive #506

### LINKED ISSUE #506: Fix request: AvoidInfiniteRecursion - this.field false positive
```java
    private Object getFieldValue(int index) {
        final String fieldName = this.getFieldNameAt(index);
        final Object fieldValueObject = this.getFieldValue(fieldName); // <- false positive 
   }
```

jborgers:


stokpop (approved):
Admire your perseverance with XPath! 

So this cannot be extended to X params using some kind of parameter/loop construct?

Is it an option to request a construct in PMD itself for cases like these? Maybe a request for a new pmd-java:checkParam() fuction or extra property on certain AST entries?

jborgers:
@stokpop Thx :-)
Maybe, would be e next step
Yes, @Type based on typeIs() would be great

PR #517, opened 2025-02-06
Fix: AvoidParallelFlux missed case #516

### LINKED ISSUE #516: Fix Request: AvoidParallelFlux missed case
Missed when no arguments to parallel():
```
private Flux<Response> processRequests(List<Request> requests, Function<Request, Response> processor) {
        return Flux.fromIterable(requests)
                .parallel() // <- false positive
                .runOn(schedulerDispatcher)
                ....
```

jborgers:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidParallelFlux.xml
```diff
@@ -52,4 +52,33 @@ class Account {
...
+import java.util.*;
+import reactor.core.publisher.Flux;
+import reactor.core.scheduler.Scheduler;
+
+class FooBad {
+    private final Scheduler scheduler;
+    private final SchedulerProperties schedulerProperties;
+    public Flux<Account> getResponseAccounts(List<AccountDetails> accountKeys, List<FieldName> requestedFields) {
+        return Flux.fromIterable(accountKeys)
+                .parallel() // bad - default parallelism - missed case
```
stokpop, line 70: I see there is also: [parallel](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#parallel-int-int-)(int parallelism, int prefetch) with three arguments, guess that will be matched as well now, which is good?

stokpop (approved):

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidParallelFlux.xml
jborgers: ah, good. Yes, I believe that is good. Never seen it used btw.

PR #523, opened 2025-02-07
Fix: false positive of UnconditionalOperationOnLogArgument #522

### LINKED ISSUE #522: Fix Request: false positive of UnconditionalOperationOnLogArgument
a debugLog method is only called conditionally

```
   private void debugLog(HttpServletRequest request, ResponseWrapper responseWrapper) throws IOException {
        // log request body after setting and before removing debugSession mdc tag
        log.debug("request headers/body request: {} / {}", getHeadersString(request), request.getInputStream() != null ? new String(IOUtils.toByteArray(request.getInputStream()), StandardCharsets.UTF_8) : "n/a");  // <---- false positive
        // log response body
        String responseBody = new String(responseWrapper.getContentAsByteArray(), StandardCharsets.UTF_8);
        String[] parts = SPLIT_PATTERN.split(responseBody);
        for (int i = 0; i < parts.length; i++) {
            log.debug("response body part {}/{}: {}", i + 1, parts.length, parts[i]); 
        }
    }

//usage:

@Slf4j
public class ResponseLogger {

    public foo() {
         if (log.isDebugEnabled()) {
                // log request headers/body and response body after setting and before removing debugSession mdc tag, use traceId and spanId to link to incidents
                debugLog(request, responseWrapper);
         }
    }
}
```

jborgers:


src/main/resources/category/java/common.xml
```diff
@@ -855,9 +855,9 @@ class Foo {
     (: if methods are chained, only report 1 in the chain :)
     and not(ancestor::MethodCall[ArgumentList[@Size>0]]/MethodCall)
   ]
-  (: except if method name starts with 'log', and it is non-public, .. :)
+  (: except if method name starts with 'log', or has 'Log' in the name, and it is non-public, .. :)
```
stokpop: comment can be: "except if method name contains log, trace, debug or info and is non-public"

stokpop (approved):
maybe update comment to reflect change?

src/main/resources/category/java/common.xml
jborgers: indeed

PR #524, opened 2025-02-07
Fix: AvoidInfiniteRecursion false positive and undetected List.of #521

### LINKED ISSUE #521: Fix request: AvoidInfiniteRecursion false positive and undetected List.of
```java
public class Foo {

    public static void applyFieldDeclaration(List<FieldDeclaration> fieldDeclarations, Object target, boolean throwExceptionOnMissingField) {
        if (fieldDeclarations != null) {
            for (FieldDeclaration declaration : fieldDeclarations) {
                applyFieldDeclaration(declaration, target, throwExceptionOnMissingField); // <- false positive, but only with "bad, detected" line
            }
        }
        applyFieldDeclaration(List.of(new FieldDeclaration("name", "value")), target, throwExceptionOnMissingField); // <- not detected
        applyFieldDeclaration(fieldDeclarations, target, throwExceptionOnMissingField); // bad, detected 
    }

    public static void applyFieldDeclaration(FieldDeclaration declaration, Object target, boolean throwExceptionOnMissingField) {
        invokeSetterOrField(target, declaration.getName(), declaration.getValue(), throwExceptionOnMissingField);
    }

    private static void invokeSetterOrField(Object target, String name, Object value, boolean throwExceptionOnMissingField) {
        // dummy
    }

    private static class FieldDeclaration {
        private final String name;
        private final Object value;

        public FieldDeclaration(String name, Object value) {
            this.name = name;
            this.value = value;
        }

        public String getName() {
            return name;
        }

        public Object getValue() {
            return value;
        }
    }
}

### PR DESCRIPTION
**Pull request by jborgers**

### CODE REVIEW

File: rulesets/java/jpinpoint-rules.xml [hunk 1]
```diff
@@ -326,12 +326,13 @@ class Good {
 (: only if the number of parameters match: not an overloaded method with difference in numbers :)
 [ArgumentList/@Size = ancestor::MethodDeclaration/FormalParameters/@Size]
 (: not if any parameter is used in the call in another position :)
-[not(ArgumentList/VariableAccess[((position() != index-of(ancestor::MethodDeclaration/FormalParameters/FormalParameter/VariableId/@Name, @Name)))])]
+[not((ArgumentList/*[self::VariableAccess and ((position() != index-of(ancestor::MethodDeclaration/FormalParameters/FormalParameter/VariableId/@Name, @Image)))]))]
```
stokpop: @Image use is deprecated in PMD 7? [See](https://pmd.github.io/pmd/pmd_userdocs_migrating_to_pmd7.html#general-ast-changes-to-avoid-image): "For PMD 7, most languages have been adapted. And when writing XPath rules, you need to replace @Image with whatever is appropriate now (e.g. @Name)."

stokpop (approved):
One question about @Image use: no alternative possible?

rulesets/java/jpinpoint-rules.xml
jborgers: Right, seems to be possible to use concat(Element/@Name, '') to make sure there is at an empty value, otherwise I get an xpath error

stokpop: hmm, at least no call to @image, but also a bit like a workaround

PR #526, opened 2025-02-10
Fix: Missed cases for AvoidMutableLists and no dealing with Lombok #525

### LINKED ISSUE #525: Fix Request: Missed cases for AvoidMutableLists and no dealing with Lombok
Example:
```
@Data
@AllArgsConstructor
@NoArgsConstructor
public class UserPojo implements Entry {

    private List<AccountPojo> accounts;
    private List<AccountError> errors;

    public UserPojo(List<AccountPojo> accounts) {
        this.accounts = accounts;
    }
}
```

The accounts field can be made final and assigned as unmodifiable in the constructor.

The field is made accessible by @ Data, with a getter and setter, so it cannot be made final. It is generally not a good idea to expose a list like this with @ Data. The list can be modified outside of the object. And that is the current way it is implemented, so I think that the cases are actually not missed, it is correct. 

We need another rule for this, AvoidExposingMutableCollections or so. There should be an explicit mutator method (addAccount) for what is needed and @ Data should be either avoided, used with @ Singular or with a unmodifiable list accounts and optionally a setter for the list (then non-final). To sort out what Lombok generates based on the field final-ness for example.

jborgers:


src/main/resources/category/java/common.xml
```diff
@@ -1461,21 +1461,26 @@ class Bad {
 
           externalInfoUrl="${doc_root}/JavaCodePerformance.md#pml01">
         <description>Problem: A list which is unnecessarily mutable may accidentally be added to and cause a memory leak. &#13;
-            Solution: Make it impossible to modify the list after object construction/initialization: use Java 9 List.of, Java 11 List.copyOf, Collections.unmodifiableList or Guava ImmutableList.
+            Solution: For the un-mutated list, make it impossible to modify the list after object construction/initialization: use Java 9 List.of, Java 11 List.copyOf, Collections.unmodifiableList or Guava ImmutableList.
```
stokpop: Bit cryptic: what is "the un-mutated list"? Not the same as "unmutable list"? Make more clear?

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidMutableLists.xml
```diff
@@ -108,4 +108,49 @@ public class Foo {
...
+    private List<AccountError> errors;
+    public UserPojo(List<AccountPojo> accounts) {
+        this.accounts = accounts;
+    }
+    public void foo() {
+        accounts.add(new AccountPojo());
+    }
+}
+
+@Data
```
stokpop, line 139: So it will be flagged when `@Value` is used? And also when no explicit contructor is present?

stokpop (approved):
See 2 remarks

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidMutableLists.xml
jborgers: @ Getter, Setter, Value and Data all make it not produce a violation, because we cannot determine if the list is actually modified

Issue #567, opened 2025-03-21
Rule request: InitializeComparatorOnlyOnce


stokpop:
Instead of using this `compareTo` implementation that creates the Comparator each invocation:

```java
@Override
public int compareTo(@NotNull Person o) {
    return Comparator.comparing(Person::getFirstName)
            .thenComparing(Person::getLastName)
            .thenComparingInt(Person::getAge)
            .compare(this, o);
}
```

initialize the comparator only once:

```java
private static final Comparator<Person> COMPARE_FIRST_LAST_NAME_AGE = Comparator.comparing(Person::getFirstName)
        .thenComparing(Person::getLastName)
        .thenComparingInt(Person::getAge);

@Override
public int compareTo(@NotNull Person o) {
    return COMPARE_AGE_LAST_FIRST.compare(this, o);
}
```

Same for these calls, but less impact:

```java
Collections.sort(persons, Comparator.comparingInt(Person::getAge).thenComparing(Person::getLastName).thenComparing(Person::getFirstName));
```

vs

```java
Collections.sort(persons, COMPARE_FIRST_LAST_NAME_AGE);
```

JMH benchmark shows:

```text
Benchmark                              Mode  Cnt         Score         Error  Units
LetsCompare.useNewComparatorPerCall   thrpt    4  17370326,029 ± 2165591,017  ops/s
LetsCompare.useSameComparator         thrpt    4  35529722,922 ± 3958701,542  ops/s

LetsCompare.useNewComparatorPerSort   thrpt    4  34628583,844 ± 2042039,234  ops/s
LetsCompare.useSameComparatorForSort  thrpt    4  41163169,536 ±  797008,887  ops/s
```

* Assume that Comparator is stateless and thread-safe. (Are there cases/examples where it is not?)

jborgers:
Rule merged so can be closed

PR #580, opened 2025-04-12
new rule #567: InitializeComparatorOnlyOnce

stokpop:
Not sure if there can be exceptions to the rule, where the Comparator used "state" that is not thread safe or can be reused.

### DIFF
--- a/docs/JavaCodePerformance.md
+++ b/docs/JavaCodePerformance.md
@@ -2654,6 +2654,75 @@ Note the use of explicit (Integer) cast to call the correct overloaded remove(In
 Suggested fix: create a category id variable of type Integer before the if statement so no further implicit boxing is done.  
 **Tip:** enable the compiler check on auto boxing/unboxing in eclipse.
 
+#### UE03
+
+**Observation: Creating new `Comparator` instances repeatedly**, e.g. in methods like `compareTo()` or collection sort operations.  
+**Problem:** Performance penalty from repeatedly creating objects, increasing garbage collection pressure and CPU usage, especially in frequently called methods or loops.  
+**Solution:** Initialize `Comparator` instances as `static final` fields. Make sure they don't use external state and are thread-safe.  
+
+Example 1: Creating a new Comparator in compareTo method:
+
+```java
+public int compareTo(Person other) {
+    return Comparator.comparing(Person::getFirstName)  // Creates new Comparator and inner lambda's each time
+            .thenComparing(Person::getLastName)
+            .compare(this, other);
+}
+```
+
+The Comparator is unnecessarily created on each method invocation. Initialize it as a static final field instead.
+
+Example 2: Proper initialization as static final field:
+
+```java
+private static final Comparator<Person> PERSON_COMPARATOR = 
+    Comparator.comparing(Person::getFirstName)
+            .thenComparing(Person::getLastName);
+
+public int compareTo(Person other) {
+    return PERSON_COMPARATOR.compare(this, other);  // Reuses existing Comparator
+}
+```
+
+Example 3: Creating a new Comparator for each TreeSet:
+```java
+public class PersonRepository {
+    public Set<Person> getPersonsSortedByName() {
+        return new TreeSet<>(Comparator.comparing(Person::getName));
+    }
+    
+    public List<Person> getSortedPersons() {
+        List<Person> persons = getPersons();
+        persons.sort(Comparator.comparing(Person::getName));
+        return persons;
+    }
+}
+```
+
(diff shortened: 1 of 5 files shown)


docs/JavaCodePerformance.md
```diff
@@ -2654,6 +2654,75 @@ Note the use of explicit (Integer) cast to call the correct overloaded remove(In
 Suggested fix: create a category id variable of type Integer before the if statement so no further implicit boxing is done.  
 **Tip:** enable the compiler check on auto boxing/unboxing in eclipse.
 
+#### UE03
+
+**Observation: Creating new `Comparator` instances repeatedly**, e.g. in methods like `compareTo()` or collection sort operations.  
+**Problem:** Performance penalty from repeatedly creating objects, increasing garbage collection pressure and CPU usage, especially in frequently called methods or loops.
```
jborgers: Can you make this into a real sentence? Like: Repeatedly creating the same object causes a performance penalty: it increases garbage collection, ...

docs/JavaCodePerformance.md
```diff
@@ -2654,6 +2654,75 @@ Note the use of explicit (Integer) cast to call the correct overloaded remove(In
 Suggested fix: create a category id variable of type Integer before the if statement so no further implicit boxing is done.  
 **Tip:** enable the compiler check on auto boxing/unboxing in eclipse.
 
+#### UE03
+
+**Observation: Creating new `Comparator` instances repeatedly**, e.g. in methods like `compareTo()` or collection sort operations.  
+**Problem:** Performance penalty from repeatedly creating objects, increasing garbage collection pressure and CPU usage, especially in frequently called methods or loops.  
+**Solution:** Initialize `Comparator` instances as `static final` fields. Make sure they don't use external state and are thread-safe.
```
jborgers: How do you mean 'external' state? Unclear. Does it mean the Comparator can have its own, internal state? 
The Comparator should not have its own mutable fields, and only use getters of the object being compared in a thread-safe way, right?

docs/JavaCodePerformance.md
```diff
@@ -2654,6 +2654,75 @@ Note the use of explicit (Integer) cast to call the correct overloaded remove(In
...
+    
+    public List<Person> getSortedPersons() {
+        List<Person> persons = getPersons();
+        persons.sort(PERSON_BY_NAME_COMPARATOR);
+        return persons;
+    }
+}
+```
+
+Note that local one-time Comparators may be acceptable in specific cases where the comparison is used only once or the logic needs to be dynamic.
```
jborgers: Can you make this into a **Note:** like elsewhere?

jborgers (changes requested):
some remarks on the doc

docs/JavaCodePerformance.md
stokpop: agree, done

docs/JavaCodePerformance.md
stokpop: yes maybe that is implicit so... removed the comment about state and thread safety

docs/JavaCodePerformance.md
stokpop: done

rulesets/java/jpinpoint-rules.xml
```diff
@@ -1757,6 +1757,67 @@ class Foo {
         </example>
     </rule>
 
+    <rule name="InitializeComparatorOnlyOnce"
+          language="java"
+          message="Avoid creating Comparator instances repeatedly. Initialize them as static final fields instead."
+          class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
+          externalInfoUrl="https://github.com/jborgers/PMD-jPinpoint-rules/tree/pmd7/docs/JavaCodePerformance.md">
+        <description>Problem: Creating Comparator instances repeatedly in methods like compareTo or sort calls is inefficient.&#13;
+            Solution: Initialize the Comparator as a static final field when not using external state and is thread safe.
```
jborgers: Removed thread safety note from doc, also remove here. I think indeed that it would be weird to put state in the comparator.

jborgers (changes requested):

rulesets/java/jpinpoint-rules.xml
stokpop: Good catch

stokpop: I also see that the externalInfoUrl can use a anchor

stokpop: done

jborgers (approved):

Issue #538, opened 2025-02-12
Fix request: check false positives due to inner classes in rules that check class fields


stokpop:
Related to this pull request: https://github.com/jborgers/PMD-jPinpoint-rules/pull/536 and issue #520 

It shows false positives for fields of inner classes. 

There might be more rules that can suffer from this.

Here the fix was: 

`ancestor::ClassBody//FieldDeclaration` -> `ancestor::ClassBody[1]/FieldDeclaration`

Investigate if more rules have this problem, a search today finds 7 uses of `ancestor::ClassBody//FieldDeclaration`

jborgers:
See also #576

jborgers:
sometimes it is:
`ancestor::ClassDeclaration//FieldDeclaration -> ancestor::ClassBody[1]/FieldDeclaration`

jborgers:
Not adding unit tests, assuming it works the same for all. Tested for #576

Issue #576, opened 2025-04-08
Fix Request: False positive in the NotProperlySynchronizingOnThisWhileUsingGuardedBy rule


Syuziko:
The variable of the violation does not have the GuardedBy annotation, yet an inner static class with a variable with the same name does have the GuardeBy annotation. So the rule gets confused.

jborgers:
It is a false positive, the rule does not differentiate between inner and outer class, both matches are accepted.
example: 
```
import net.jcip.annotations.GuardedBy;

class Issue576 {
    @GuardedBy("this")
    private String txt1;
    private String txt2;

    public void modifyTextSafe() {
        synchronized(this) {
            txt1 += "-appended";
            txt2 += "-appended";
        }
    }

    public boolean modifyTextUnSafe() {
      txt1 += "-appended"; // bad
      txt2 += "-appended"; // <-- false positive
    }

    static class InnerIssue {
        private String txt1;
        @GuardedBy("this")    
        private String txt2;

        public void modifyTextSafe() {
          synchronized(this) {
            txt1 += "-appended";
            txt2 += "-appended";
          }
      }

      public boolean modifyTextUnSafe() {
        txt1 += "-appended"; // <-- false positive
        txt2 += "-appended"; // bad
      }
    }
}

```

PR #582, opened 2025-04-14
Fix : check false positives due to inner classes in rules that check class fields #538

jborgers:


rulesets/java/jpinpoint-rules.xml
```diff
@@ -384,7 +384,7 @@ class Good {
     ]/(ClassType/@SimpleName|PrimitiveType/@Kind)
     or
     (: or its classType is equal to the field classtype :)
-    (ClassType/@SimpleName|PrimitiveType/@Kind) = ancestor::ClassBody//FieldDeclaration[
+    (ClassType/@SimpleName|PrimitiveType/@Kind) = ancestor::ClassBody[1]/FieldDeclaration[
```
stokpop, line 387: replacing the `//` with `[1]/` is ok, so FieldDeclaration is always _direct_ parent?

stokpop: maybe add one or two unit tests to prove this fix and have some regression test for future?

stokpop (changes requested):

rulesets/java/jpinpoint-rules.xml
jborgers: As far as I know, it is always a direct child.

jborgers: Assuming it works the same for all. Tested for #576, has unit test for this.

stokpop: ok, let's try it with some runs with IntelliJ PMD after merge :-)

stokpop (approved):

PR #583, opened 2025-04-14
fix for #549: add documentation about @SuppressWarning

### LINKED ISSUE #549: Fix request: example is missing for UsingSuppressWarnings
Give example how to use SuppressWarning correctly.

Also add an item for this to the markdown documentation.

stokpop:


docs/JavaCodeQuality.md
```diff
@@ -521,6 +521,58 @@ private static final int SIXTIES_START = 1960; // good
...
+It will specify the rule you want to suppress explicitly, and also ask you to specify a reason for the suppression.
+```java
+@SuppressWarnings({"pmd:AvoidConcatInAppend", "pmd:UsingSuppressWarnings"}) //NOPMD - suppressed UsingSuppressWarnings - TODO explain reason for suppression
+class Foo implements KeyGenerator {
+    // ...
+}
+```
+In this example you see that you can use it to suppress the SuppressWarnings rule itself. This is not a good idea, because it is not clear why you are suppressing the rule.
+The Intellij PMD Plugin will give issues on information level about the usage of `// NOPMD` so they can be easily reviewed.
+The `// NOSONAR` is not checked by both PMD Plugin and Sonar, so be careful with its usage.
```
jborgers: so be careful with its usage -> avoid its usage.

docs/JavaCodeQuality.md
```diff
@@ -521,6 +521,58 @@ private static final int SIXTIES_START = 1960; // good
...
+@SuppressWarnings("pmd:UnconditionalCreatedLogArguments") // Reason for suppression: the rule doesn't recognize record getters (yet), and those are not expensive
+public void process(ConsumerRecord consRecord) {
+    ConsumerEvent event = consRecord.event(); // record getter
+    log.info("Received event '{}'", event); // pmd:UnconditionalCreatedLogArguments suppressed
+}
+```
+**Note:** This rule is just informational to track use of SuppressWarnings, to be able to double-check if suppression is correct. It often doesn't need a fix.
+The @SuppressWarning is the preferred way to suppress a rule violation, because you can specify the exact rule you want to suppress. Downside is that it does
+not work on line level. Two other options are:
+* `// NOPMD`: suppresses all PMD violations on the line, not just the one you want to suppress.
```
jborgers: Note that this notation has no effect in Sonar, just PMD.

docs/JavaCodeQuality.md
```diff
@@ -521,6 +521,58 @@ private static final int SIXTIES_START = 1960; // good
...
+* `// NOSONAR`: suppresses all Sonar violations on the line, not just the one you want to suppress.
```
jborgers: Note that this notation has no effect in the PMD tool, just in Sonar.

docs/JavaCodeQuality.md
```diff
@@ -521,6 +521,58 @@ private static final int SIXTIES_START = 1960; // good
...
+### M02
+
+**Observation: @SuppressWarnings is used**  
+**Problem:** The suppressed rule detects problems, suppressing the detected violations without full knowledge can lead to the problems the rule is trying to prevent.  
+**Solution:** Only suppress violations (warnings) when you have complete understanding of the rule, and you are sure it does not apply.
+When suppressing warnings, document your reasoning and report false positives to the rule maintainers to help improve rule accuracy.  
+**Rule name:** UsingSuppressWarnings  
+**Example:**
+```java
+@SuppressWarnings("pmd:UnconditionalCreatedLogArguments") // Reason for suppression: the rule doesn't recognize record getters (yet), and those are not expensive
```
jborgers: ik geloof dat je ook **PMD**.UnconditionalCreatedLogArguments kunt doen, maar dat dat alleen in PMD werkt. Misschien even checken en ook toevoegen aan doc.

docs/JavaCodeQuality.md
```diff
@@ -521,6 +521,58 @@ private static final int SIXTIES_START = 1960; // good
...
+
+**Observation: @SuppressWarnings is used for high risk problems**  
+**Problem:** The suppressed rule detects high risk problems, suppressing the detected violations without full knowledge can lead to incidents like customer data mix-up, corrupt data, server crashes or very bad performance.  
+**Solution:** Only suppress violations (warnings) when you have complete understanding of the rule, and you are sure it does not apply.
+When suppressing warnings, document your reasoning and report false positives to the rule maintainers to help improve rule accuracy.  
+**Rule name:** UsingSuppressWarningsHighRisk  
+**Note:** This rule is to track use of SuppressWarnings for high risk rule violations, to be able to double-check if suppression is correct. It doesn't necessarily need a fix.   
+**Example:**  
+```java
+@SuppressWarnings({"pmd:AvoidUnguardedMutableFieldsInSharedObjects", "PMD.AvoidUserDataInSharedObjects"}) // suspicious, bad to suppress
```
jborgers: ah, hier `pmd: `en `PMD.` gebruik

jborgers: suspicious, bad to suppress -> suspicious, **probably** bad to suppress (also in rule description?)

jborgers (changes requested):

docs/JavaCodeQuality.md
stokpop: hmm interesting: `PMD.` works in Sonar and IntelliJ PMD plugin, `pmd:` only works in Sonar!

stokpop: And using `pmd.AvoidFutureJoinWithoutTimeout` works also in Sonar, but also seems to trigger "Using SuppressWarnings for a rule that is meant to prevent high risk problems."?

stokpop: Can it trigger this regex? `PMD[^\.]|pmd[^:]` from:

```xml
<property name="ruleIdMatches" type="String" value="AvoidUserDataInSharedObjects|AvoidUnguardedMutableFieldsInSharedObjects|AvoidUnguardedAssignmentToNonFinalFieldsInSharedObjects|AvoidMutableStaticFields|[^\w]ALL[^\w]|[^\w]all[^\w]|PMD[^\.]|pmd[^:]"
           description="Regex for inclusion of high risk rules"/>
```

stokpop: note that when using `PMD.` there is also a list of "Suppressed violations by Annotation" in IntelliJ plugin report

jborgers: ah, I didn't know about `pmd.`. Right, in the regex it is not one of those.

jborgers: I didn't know `PMD.` also works in sonar. So, that would be the best way, right? both tools supported, PMD shows suppressedViolations. Does it not do that with `pmd:`?

stokpop: no, there is no "suppressed violations" with `pmd:`, and also the suppress itself does not work in PMD... there is only the "UsingSuppressWarning"... so `PMD.` seems the best option now

stokpop: what is reason of `PMD[^\.]|pmd[^:]`? -> seems: using prefix of PMD or pmd without . or : respectively will trigger the `UsingSuppressWarningsHighRisk` rule? why?

jborgers: No, de `[^..]` means a **not** any of the characters, so it allows for `PMD.` and `pmd:`, and not for instance `PMDALL` or so (if it would be something with a meaning).

stokpop: Ah, found it in unit test, and added this description: 

The `@SuppressWarnings("pmd")` or `@SuppressWarnings("PMD")` are considered high risk as it suppresses many rules at once, use explicit rule list instead.

stokpop: should we fix the UsingSuppressWarningsHighRisk regex to not match on `pmd.` and `PMD:`?

jborgers: Nice to have documentation in the unit tests :-)

jborgers: Well, pmd. and PMD: does not introduce high risk, only wrong suppression, which won't work, so they will fix it by them selves, to make it suppress, right?

jborgers (approved):

jborgers:
@stokpop I think you only have to resolve the conversation

docs/JavaCodeQuality.md
stokpop: Yes, I would think so. Let's see how it works out.

PR #595, opened 2025-04-28
New rule: Avoid wrong use of @GuardedBy; upgrade to pmd 7.12

### LINKED ISSUE #537: Rule Request: Avoid wrong use of @GuardedBy
@ GuardedBy is used is wrong apparently to get rid of violations which indicate concurrency bugs, so high risk. Related to rule: NotProperlySynchronizingOnFieldWhileUsingGuardedBy

Example 1:
```
class Foo {
   @GuardedBy("lock")
    private Map<String, Product> cachedData = new HashMap<>();
}
```
And no `lock` field exists.

Example 2:

```
@Component
@ConfigurationProperties(prefix = "my-application")
@Getter
@Setter
public class MyApplicationConfig {
    
    @GuardedBy("this")
    private String bannerText;
    @GuardedBy("this")
    private String buttonText;

}
```
There is no synchronized on "this" involved. Used to get rid of the violation, it this case because of the `@Component` annotation. Proper fix is to make it `@Configuration` instead of `@Component`.

jborgers:


rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -4806,6 +4806,69 @@ public class Foo {
...
+)][@Name = ancestor::ClassBody[1]//FieldDeclaration/ModifierList/Annotation[@SimpleName='GuardedBy']//MemberValuePair/StringLiteral/@ConstValue]
+                ]]></value>
+            </property>
+        </properties>
+        <example><![CDATA[
+class WrongUseOfGuardedByExample {
+    public static final String FIELD = "field"; // bad: should be 'private' and initialized with 'new Object()' or 'new Object[0]'
+
+    @GuardedBy(FIELD) // bad, not a String literal
+    private static String txt;
```
stokpop: should a static String field like this really be guarded? Maybe a "stronger" example when using a real mutable object such as plain HashMap?

docs/JavaCodePerformance.md
```diff
@@ -2521,6 +2521,73 @@ class SingletonBMGood {
 }
 ```
 
+#### TUTC14
+**Observation: Locking and describing the locking behaviour with `@GuardedBy` is applied in a wrong way or not meeting best practices.**   
+**Problem:** Wrong use of locking and `@GuardedBy` may result in thread-unsafety and concurrency bugs. It may obstruct other GuardedBy rules which check whether the promise of guarding is actually met, it may disguise high risk violations.  
+**Solution:** Use `@GuardedBy` properly: 
+* use a private final lock object constructed with `new Object()` or `new Object[0]`
```
stokpop, line 2528: what about using `java.util.concurrent.locks.*Lock` variants? Use of these will cause false positives?

stokpop (changes requested):

docs/JavaCodePerformance.md
jborgers: Those are not supported, yet. Don't know what will happen.

rulesets/java/jpinpoint-java-rules.xml
jborgers: Right, the original problem was a HashMap, this is a (too) simplified test case. Will change.

docs/JavaCodePerformance.md
stokpop: OK, let's put this in scope of another issue/fix request when the need arrises.

stokpop (approved):

PR #598, opened 2025-05-19
Rule request: HttpInterceptorNotReleasingOnException - ClientHttpRequestInterceptor: should release resources if exception is thrown #594

### LINKED ISSUE #594: Rule request: HttpInterceptorNotReleasingOnException - ClientHttpRequestInterceptor: should release resources if exception is thrown
Found this to cause issues during several troubleshoots already. If an exception is thrown in a ClientHttpRequestInterceptor, resources must be released by calling ClientHttpResponse.close(). Failing to do so causes connections not to be released to the connection pool for instance.

As per documentation of ClientHttpRequestInterceptor.intercept():
> Note: if the interceptor throws an exception after receiving a response, it must close the response via ClientHttpResponse.close().

Example:
```java
public class ValidatingClientHttpRequestInterceptor implements ClientHttpRequestInterceptor {
    @Override
    public ClientHttpResponse intercept(HttpRequest request, byte[] body, ClientHttpRequestExecution execution) throws IOException {
        final ClientHttpResponse response = execution.execute(request, body);

        // should call ClientHttpResponse.close() and release resources if an exception is thrown
        someValidation(response);

        return response;
    }

    public static void someValidation(ClientHttpResponse response) {
        if (!response.getHeaders().containsKey("X-Some-Header")) {
            throw new RuntimeException("some error");
        }
    }
}
```

jborgers:


rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -6110,6 +6110,61 @@ public class HttpClientStuff {
...
 
+    <rule name="HttpInterceptorNotReleasingOnException"
+          message="ClientHttpRequestInterceptor is not releasing the connection when it throws an Exception."
+          class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
+          language="java"
+          externalInfoUrl="https://github.com/jborgers/PMD-jPinpoint-rules/tree/pmd7/docs/JavaCodePerformance.md#ibi25">
+        <description>Problem: If the interceptor throws an exception after receiving a response, resources are not released as happens with normal program flow.
+            It causes the connection not to be released to the connection pool, which leads to pool exhaustion and unresponsiveness.
+            Solution: Release resources by closing the response via ClientHttpResponse.close() when throwing an Exception.
+            (jpinpoint-rules)
```
stokpop, line 6121: why both jpinpoint-rules and jpinpoint-java-rules? only the latter, right?

rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -6110,6 +6110,61 @@ public class HttpClientStuff {
...
+    if (!response.getHeaders().containsKey("X-Some-Header")) {
+      // Log error
+      throw new MyException("validation error"); // bad, close missing
+    }
+  }
+
+  private static void someValidationGood(ClientHttpResponse response) throws MyException {
+    if (!response.getHeaders().containsKey("X-Some-Header")) {
+      // log error
+      response.close(); // good
```
stokpop, line 6159: Feels like it should not be the responsibility of this method to close the connection... you can also lift this to the caller method that created the response with a try / catch MyException (or some parent exception class) to close the connection? (are both recognized by the rule?)

docs/JavaCodePerformance.md
```diff
@@ -581,6 +581,42 @@ public class Foo {
...
+  private static void someValidationGood(ClientHttpResponse response) throws MyException {
+    if (!response.getHeaders().containsKey("X-Some-Header")) {
+      // log error
+      response.close(); // good
+      throw new MyException("validation error");
+    }
+  }
+}
+```
+**See:** [ClientHttpRequestInterceptor javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/ClientHttpRequestInterceptor.html)
```
stokpop, line 633: Good to have this link! There is an explicit note added indeed via [this commit](https://github.com/spring-projects/spring-framework/commit/0542fe5232a54bc7f92c5ee43856ebe4acd5d835). I was looking at a project with older SpringBoot version without this note in the javadoc...

rulesets/java/jpinpoint-java-rules.xml
stokpop: I see there is a unit test doing this below :-) Should that case be promoted to example, that solution might be prefered considering single-responsibility on method level?

stokpop (changes requested):
approved, but with some questions/remarks that might need small changes

rulesets/java/jpinpoint-java-rules.xml
jborgers: You mean `(jpinpoint-rules)`? That is consistent with the rest of the source file (remoting.xml). 
I am not sure where the renaming responsibility lays, with the renaming and copying from jpinpoint-rules.xml to jpinpoint-java-rules.xml. At least it should be consistent, I see it is not consistent now.

rulesets/java/jpinpoint-java-rules.xml
jborgers: I think it indeed is a better quality solution, I chose for the shorter example. I will change it to the better solution.

jborgers: made it short in the rule definition and longer on the doc page

rulesets/java/jpinpoint-java-rules.xml
stokpop: but now they are both present, see line 6122, should that one be removed?

jborgers: as discussed, the source remoting.xml is fine. The merge tool causes this with the new filename. To be dealt with in a separate issue.

PR #603, opened 2025-05-30
fix for #373: fix xpath and added dropped support warnings for portlets in Spring

### LINKED ISSUE #373: Fix request (PMD7): AvoidModelMapAsRenderParameter - improve unit tests coverage - fix assumed false positives
As found during PMD6 to PMD7 migration:

* unit tests not complete: non-matches are missing, second part of xpath expression is not covered --> add unit tests
* investigate and fix for comment for two bad ones: "[--> NOT, see JPCC-88?]" --> the Jira item is gone, check the issue and fix accordingly

stokpop:
closes #373 

created "band-aid" solution, should we keep this rule for dropped portlet support?

### DIFF
--- a/docs/JavaCodePerformance.md
+++ b/docs/JavaCodePerformance.md
@@ -1458,6 +1458,9 @@ Note that for StringBuilder.substring this is not needed, since it does make a c
 
 #### TMSU11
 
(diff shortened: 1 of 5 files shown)


jborgers (approved):
Bandaid fix is fine by me. People might still use it, I believe we should not yet drop it.

PR #608, opened 2025-06-13
fix #607: Kotlin: also have AvoidDecimalAndChoiceFormatAsField work in object instead of class

### LINKED ISSUE #607: Fix request: Kotlin PMD detects issues in class, but not object
Hi,

When executing the PMD plugin check within IntelliJ we're finding that the rule violations aren't being detected in some of the class structures.

With the following structure it works fine:
Standard class structure
```
class DecimalFunctionClass {
    val df: DecimalFormat = DecimalFormat("#.###")

    fun format(value: Double) = df.format(value)
}
```

However in the following structure it doesn't find the violation:
Object class structure
```
object DecimalFunctionClass {
    val df: DecimalFormat = DecimalFormat("#.###")

    fun format(value: Double) = df.format(value)
}
```

We're listing this as a fix request as it can be something used by teams and might miss some of the violations

stokpop:
Not sure what fix is better:

`//(ClassDeclaration|ObjectDeclaration)`

or 

`//TopLevelObject` (this is now in the rule)

### DIFF
--- a/src/main/resources/category/kotlin/common.xml
+++ b/src/main/resources/category/kotlin/common.xml
@@ -94,7 +94,7 @@ class AvoidStringBuffer {
 //ImportHeader[.//T-Identifier[@Text='java'] and .//T-Identifier[@Text='text']][1]/ancestor::KotlinFile
 (: PrimaryExpression finds the constructors instead of type declarations :)
 (: filter matches in functions :)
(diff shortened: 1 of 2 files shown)


stokpop:
Still checking if other rules need updates that use `ClassDeclaration` somewhere... have not found other issues yet, and other rules also have test cases with "object"

jborgers (approved):

Issue #610, opened 2025-06-20
Utilize md layout for enumerations


jborgers:
Make enumerations look nice.

Issue #605, opened 2025-06-04
Utilize md layout for rule documentation in IntelliJ plugin


jborgers:
IntelliJ plugin with the new PR https://github.com/amitdev/PMD-Intellij/pull/246 supports markdown in rule documentation.
The doc needs to utilize this, for instance when for numbered lists not utilized: 

![Image](https://github.com/user-attachments/assets/2af306bd-3dd6-445a-9192-97ab4b169654)

When utilized:

![Image](https://github.com/user-attachments/assets/4c2d544b-ae11-4a6c-8b15-ccd8cd354c50)

Same holds for instance for backtick (`) usage for code.  

Achieved in concurrent.xml with:

<img width="465" alt="Image" src="https://github.com/user-attachments/assets/7a924821-4967-454e-9e0f-c465ba6d44a7" />

PR #611, opened 2025-06-20
Utilize md layout for enumerations #610 (sub task of  #605)

jborgers:


stokpop (commented):
I assume these are all occurrences of enumerations

stokpop (approved):

PR #613, opened 2025-06-23
Fix: false positive in BufferingClientHttpRequestFactoryIsMemoryGreedy #596

### LINKED ISSUE #596: Fix Request: false positive in BufferingClientHttpRequestFactoryIsMemoryGreedy
https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-6.1-Release-Notes#web-applications
 
To reduce memory usage in RestClient and RestTemplate, most ClientHttpRequestFactory implementations no longer buffer request bodies before sending them to the server. As a result, for certain content types such as JSON, the contents size is no longer known, and a Content-Length header is no longer set. If you would like to buffer request bodies like before, simply wrap the ClientHttpRequestFactory you are using in a BufferingClientHttpRequestFactory.

Reported: For our service we need a connection with Azure hsm, which always needs a content length header. So buffering is needed.

Questions: is this often the case? Do we still want to apply this rule? Can we make exceptions on the rule? Or should we make it lower (info?) severity and make the programmer aware and consider it,  find out if needed or not?

At least the documentation needs to be updated to explain this.

jborgers:
For #596

### DIFF
--- a/docs/JavaCodePerformance.md
+++ b/docs/JavaCodePerformance.md
@@ -392,8 +392,12 @@ class MyClientHttpRequestFactorySupplier implements Supplier<ClientHttpRequestFa
 
 #### IBI18
 **Observation: Spring BufferingClientHttpRequestFactory is used.**   
-**Problem:** org.springframework.http.client.BufferingClientHttpRequestFactory buffers all incoming and outgoing streams fully in memory which may result in high memory usage.   
-**Solution:** Avoid multiple reads of the response body so it is not needed.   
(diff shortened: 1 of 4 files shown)


stokpop (commented):
So that means that the receiving http server does not accept `Transfer-Encoding:chunked`? 

Chucked is preferred for larger and dynamically generated content.

So how often is this an issue? Often enough to make this `INFO` level instead of `WARN` with an explicit `SuppressWarning`?

stokpop (dismissed):

PR #620, opened 2025-07-14
fix #618: skip Set<Class> in formal parameters as cannot tell instantatiation of Set

### LINKED ISSUE #618: Fix Request: false positive in NonComparableSetKeys for Java
```
public class Foo {

    public static final Comparator<Class> CLASS_NAME_COMPARATOR = Comparator.comparing(Class::getName);

    public void bar() {
        final Set<Class> types = new TreeSet<>(CLASS_NAME_COMPARATOR);
        foo(types);
    }

    public void foo(Set<Class> types) { // <--- false positive, has constructor with comparator
        types.contains(String.class);
    }

}
```

Should we check all type declarations? In case above we can check the TreeSet construction in same class, but that is not always the case.

The constructor itself is also flagged, so maybe not flag declarations? 

(Same for NonComparableMapKeys rule?)

stokpop:
To prevent false positives the matches on FormalParams has been removed: you cannot tell the Set implementation and initiation (expect in same class, but that is already checked by the rule).

Question: should we add to documentation to use SuppressWarning annotation when it is known the set is only for small numbers and the risk of having more than 7 elements in a _bucket_ is minimal?

### DIFF
--- a/src/main/resources/category/java/common.xml
+++ b/src/main/resources/category/java/common.xml
@@ -1971,10 +1971,10 @@ public enum Fruit {
 (: declared variable used in Set by-element access :)
 //VariableId[@Name = ancestor::ClassBody//MethodCall[@MethodName = ('contains','containsAll','remove','removeAll','retainAll')]/VariableAccess/@Name]
     (: for 3 ways defined as a Set :)
-    [(ancestor::FieldDeclaration|ancestor::LocalVariableDeclaration|ancestor::FormalParameter)/ClassType[pmd-java:typeIs('java.util.Set')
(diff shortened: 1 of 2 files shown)


jborgers (approved):

jborgers:
> To prevent false positives the matches on FormalParams has been removed: you cannot tell the Set implementation and initiation (expect in same class, but that is already checked by the rule).
> 
> Question: should we add to documentation to use SuppressWarning annotation when it is known the set is only for small numbers and the risk of having more than 7 elements in a _bucket_ is minimal?

We could add that indeed, as an option to add the suppression, in case that is known. Of course over time there might be the risk that number of elements will grow.

PR #621, opened 2025-07-15
fix #618: improve documentation and tests, merge and added ConcurrentSkipListSet

### LINKED ISSUE #618: Fix Request: false positive in NonComparableSetKeys for Java

stokpop:


src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/NonComparableSetElements.xml
```diff
@@ -5,33 +5,37 @@
...
-    Set<Thread> threadFieldSet = new HashSet<>(); // bad 6
+    Set<Thread> threadFieldSet = new HashSet<>(); // bad 8
     Thread t = new Thread();
     List<String> strList = new ArrayList<>();
     List<Thread> tList = new ArrayList<>();
-    Set<HttpRoute> routeSet = new HashSet<>();
+    Set<HttpRoute> routeSet = new HashSet<>(); // bad 12, HttpRoute not comparable
     Set tSet = new HashSet();
+    Set<HttpRoute> fieldRouteSetNonComp = new TreeSet<>(); // bad 14, HttpRoute not comparable and no explicit comparator
+    Set<HttpRoute> fieldRouteSetConc = new ConcurrentSkipListSet<>(); // bad 15: HttpRoute not comparabl with explicit comparator
```
jborgers: comparabl -> comparable

jborgers (dismissed):

jborgers (approved):

PR #631, opened 2025-08-28
Fix: false positive in AvoidTimeUnitConfusion #630

### LINKED ISSUE #630: Fix Request: false positive in AvoidTimeUnitConfusion
When type is Duration, it knows about the time unit, so no violation.
```
class SomeListener {
   public SomeListener(@Value("${expire-duration}") Duration expireDuration) { // <-- false positive
      this.expireDuration = expireDuration;
   }
}
```

jborgers:


rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -1058,7 +1058,7 @@ or ends-with(lower-case(@Name), 'ttl')
 or ends-with(lower-case(@Name), 'timestamp')]
 ,
 (: @Value :)
-//Annotation[@SimpleName='Value' and .//MemberValuePair[@Image='value']/StringLiteral[
+//Annotation[../../PrimitiveType and pmd-java:typeIs('org.springframework.beans.factory.annotation.Value') and .//MemberValuePair[@Image='value']/StringLiteral[
```
stokpop: Assuming the PrimitiveType will skip a type like Duration: what if it is of non primitive type but still without time unit indication, e.g. `java.lang.Long`?

stokpop (changes requested):

rulesets/java/jpinpoint-java-rules.xml
jborgers: Will allow Long and Integer

stokpop (approved):

PR #650, opened 2025-10-07
New Rule: move from Joda time and threeten.bp to java.time #622

### LINKED ISSUE #622: Rule Request: move from Joda time and threeten.bp to java.time
Several projects still use Joda time.

> The standard date and time classes prior to Java SE 8 are poor. By tackling this problem head-on, Joda-Time became the de facto standard date and time library for Java prior to Java SE 8. Note that from Java SE 8 onwards, users are asked to migrate to java.time (JSR-310) - a core part of the JDK which replaces this project.

LocalDateTime memory usage:
Joda-time: 80 bytes 
java.time: 48 bytes

So Joda time takes 67% more memory. More memory usage also means more processing time.

This helps for performance, sustainability and to save on cloud costs.

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -3228,6 +3228,25 @@ Use of slow library calls
...
+#### UOSLC03
+
+**Observation: Legacy time library: Joda-time or ThreeTenBp is used.**   
+**Problem:** These legacy libraries have non-optimal performance and memory usage.   
+**Solution:** Migrate to java.time for a modern solution with better performance and less memory usage.   
+**Rule name:** LegacyTimeLibraryUsed
+
+**Indicative Performance & Memory Differences: Joda-Time vs ThreeTenBP vs java.time**
+
+| **Operation Type**     | **Joda-Time**                                 | **ThreeTenBP**                              | **`java.time` (Java 8+)**              |
```
stokpop: Is there a source for the data in this table?

rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -1830,6 +1830,34 @@ public class GoodPerson implements Comparable<Person> {
         </example>
     </rule>
 
+    <rule name="LegacyTimeLibraryUsed"
+          language="java"
+          message="Migrate from the legacy time library to java.time for better performance."
+          class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
+          externalInfoUrl="https://github.com/jborgers/PMD-jPinpoint-rules/tree/pmd7/docs/JavaCodePerformance.md#uoslc03">
+        <description>Problem: the JodaTime and ThreeTenBp libraries have non-optimal performance and memory usage.
```
stokpop, line 1838: The ThreeTen backport release is created for java 6 and 7 projects. Mention this? There is no alternative for these so should be suppressed? Can suppression happen on import statements? 

Same for JodaTime usage in pre java 8?

rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -1830,6 +1830,34 @@ public class GoodPerson implements Comparable<Person> {
...
+            Solution: Migrate to java.time for a modern solution with better performance and less memory usage.
+        (jpinpoint-rules)</description>
+        <priority>2</priority>
+        <properties>
+            <property name="tags" value="jpinpoint-rule,performance,sustainability-medium,cpu,memory" type="String" description="classification"/>
+            <property name="xpath">
+                <value><![CDATA[
+//ImportDeclaration[starts-with(@PackageName, 'org.joda.time')][1]
+|
+//ImportDeclaration[starts-with(@PackageName, 'org.threeten.bp')][1]
```
stokpop, line 1848: Can there also be cases of `org.threeten` before the backport `bp`?

rulesets/java/jpinpoint-java-rules.xml
```diff
@@ -1830,6 +1830,34 @@ public class GoodPerson implements Comparable<Person> {
...
+          externalInfoUrl="https://github.com/jborgers/PMD-jPinpoint-rules/tree/pmd7/docs/JavaCodePerformance.md#uoslc03">
+        <description>Problem: the JodaTime and ThreeTenBp libraries have non-optimal performance and memory usage.
+            Solution: Migrate to java.time for a modern solution with better performance and less memory usage.
+        (jpinpoint-rules)</description>
+        <priority>2</priority>
+        <properties>
+            <property name="tags" value="jpinpoint-rule,performance,sustainability-medium,cpu,memory" type="String" description="classification"/>
+            <property name="xpath">
+                <value><![CDATA[
+//ImportDeclaration[starts-with(@PackageName, 'org.joda.time')][1]
```
stokpop, line 1846: Considered matching on line where used instead?

stokpop (changes requested):

docs/JavaCodePerformance.md
jborgers: Copilot :-) mention it?

rulesets/java/jpinpoint-java-rules.xml
jborgers: Right, good to mention. Suppression not with annotation but with // NOSONAR // NOPMD is possible as far as I know

rulesets/java/jpinpoint-java-rules.xml
jborgers: I didn't find.

rulesets/java/jpinpoint-java-rules.xml
jborgers: I considered that, yet, the typeIs is more brittle and I assume taking more resources. This is more simple, also the rule expression.

docs/JavaCodePerformance.md
stokpop: I have some doubt about the contents of the table. For instance, it mentions "Not all components are thread-safe" for Joda time, but that seems not fair. 

The website of Joda time says: "Every public class in Joda-Time is documented as being thread-safe or not. Joda-Time makes heavy use of the immutability design pattern, and all immutable classes in Joda-Time are thread-safe. Many mutable classes whose instances are not likely to be shared are not thread-safe, and are documented as such." (https://www.joda.org/joda-time/faq.html#threading). 

So what about the other fields, e.g percentages and claims about object creation and wrappers? Maybe make more general statement? Or ask copilot for the sources. :-)

stokpop (commented):
Have some doubt about the (use of) ai comparison table. See comment.

docs/JavaCodePerformance.md
jborgers: Good catch! Never trust AI!
I really needed to ask to double check and show the documentation. I checked with Gemini and Grok, both made the same mistake. I also asked Grok how this mistake can be explained, also for all three, see: https://grok.com/share/c2hhcmQtMg%3D%3D_8f52aaa9-8b5b-4482-9a1e-b06b5219248d

stokpop (approved):

PR #660, opened 2025-11-06
Fix: use of ReentrantLock in NotProperlySynchronizingOnFieldWhileUsingGuardedBy #474

### LINKED ISSUE #474: Fix request: use of ReentrantLock in NotProperlySynchronizingOnFieldWhileUsingGuardedBy
Rule does not detect use of ReentrantLock. For example:

```java
import org.springframework.stereotype.Component;
import javax.annotation.concurrent.GuardedBy;
import java.util.concurrent.locks.Lock;
import java.util.concurrent.locks.ReentrantLock;
import java.util.regex.Pattern;

@Component
public class Foo {
    private final Lock lock = new ReentrantLock();

    @GuardedBy("lock")
    private Pattern pattern;
    
    public void doChangePattern(String text) {
        lock.lock();
        try {
            pattern = Pattern.compile(text);
        }
        finally {
            lock.unlock();
        }
    }
}
```

Possible way for solution: check for use of `lock.lock()` outside of surrounding try/finally, of typeIs `java.util.concurrent.locks.Lock`?

Should also added to other rules: `NotProperlySynchronizingOnFieldWhileUsingGuardedBy`, ...

Workaround is to use old fashioned Object lock with `synchronized` keywords. (But might not work well with virtual thread pinning.)

jborgers:


stokpop (changes requested):
Also update the description with use of ReentrantLock?

Now it mentions: `synchronize access by a synchronized(LOCK) block`

docs/JavaCodePerformance.md
```diff
@@ -2230,7 +2230,12 @@ public boolean equals(final Object arg0) {
 Thread unsafety and lock contention
 -----------------------------------
 
-When multiple threads access the same object, access it in a thread safe way. Getting thread safety right and not hindering performance is difficult. Locking with the synchronized keyword may introduce lock contention under load, which is bad for performance. To make threading aspects easier to understand in source code, we recommend the use of [Java Concurrency In Practice annotations.](http://jcip.net.s3-website-us-east-1.amazonaws.com/annotations/doc/index.html)
+When multiple threads access the same object, access it in a thread safe way. Getting thread safety right and not hindering performance is difficult. 
+Locking with the synchronized keyword may introduce lock contention under load, which is bad for performance. To make threading aspects easier to understand in source code, we recommend the use of [Java Concurrency In Practice annotations.](http://jcip.net.s3-website-us-east-1.amazonaws.com/annotations/doc/index.html)
```
stokpop: Secure direct link: https://jcip.net/annotations/doc/index.html

docs/JavaCodePerformance.md
```diff
@@ -2230,7 +2230,12 @@ public boolean equals(final Object arg0) {
...
+
+### Virtual threads and locking
+Virtual threads are a very efficient way to handle concurrency, especially for I/O bound tasks. Up to JDK 24, they have an issue with doing I/O within a synchronized block, called thread pinning.
+To work around this, use a Reentrant lock with try-finally. However, we consider it risky and do *not* recommend to use virtual threads before JDK 24. Therefore, we also do not promote this workaround.
```
stokpop: why risky? due to the thread pinning, right? not the use of Reentrant lock per se.

stokpop (dismissed):
Can you update the link? And maybe add some clarification of "risky"?

docs/JavaCodePerformance.md
jborgers: Nice, replaced

docs/JavaCodePerformance.md
jborgers: Ah, unclear. Rewritten.

stokpop (approved):

Issue #534, opened 2025-02-11
Rule Request: AvoidNonAtomicIfModifyOnConcurrentMap


jborgers:
Check-Modify constructs are the most subtle and most occurring concurrency bugs. When using a ConcurrentMap, it is a multi-threading environment, so separate check and modify is a concurrency bug.

Example:

```
void foo(ConcurrentMap<String, Account> accountMap, String accKey) {
     if(!accountMap.containsKey(accKey)){
            accountMap.put(accKey, account);
      }
}
```

`accountMap.putIfAbsent(accKey, account)` should be used

jborgers:
Note it needs support for local var (used in parallelStream), formal param and field.

jborgers:
Not just put: 
Utilize an atomic if-combined-with-modify operation provided by the `ConcurrentMap`: `putIfAbsent`, `computeIfAbsent`, `computeIfPresent`, `getOrDefault`, `remove` and `replace`.

PR #664, opened 2025-11-12
New Rule: AvoidNonAtomicIfModifyOnConcurrentMap #534

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -2681,6 +2681,38 @@ class LombokInstanceLockUsedGood {
...
+            accountMap.put(accKey, account);
+        }
+    }
+
+    void goodPut(ConcurrentMap<String, String> accountMap, String accKey, String account) {
+        accountMap.putIfAbsent(accKey, account);
+    }
+
+    void badRemove(ConcurrentMap<String, String> accountMap, String accKey, String account) {
+        if (accountMap.containsKey(accKey) && Objects.equals(accountMap.get(accKey), account)) {
```
stokpop: get will return null when accKey does not exist, so probably only the Objects.equals is needed because it can deal with null values?

docs/JavaCodePerformance.md
```diff
@@ -2681,6 +2681,38 @@ class LombokInstanceLockUsedGood {
...
+A ConcurrentMap is used in a multi-threading environment, and a separate if and modify is a concurrency bug because one thread can execute the if operation, be scheduled-out, a second thread also executes the if operation, and then both will do the modify operation. The if and modify need to be atomically combined.   
+**Solution:** Utilize an atomic if-combined-with-modify operation provided by the `ConcurrentMap`: `putIfAbsent`, `computeIfAbsent`, `computeIfPresent`, `getOrDefault`, `remove` and `replace`.   
+**Note:** A `get` is not a modify, so if-get is stricly not a bug. Yet, replacing it with the atomic `getOrDefault` is recommended,   
+**Rule name:** AvoidNonAtomicIfModifyOnConcurrentMap.   
+**See:** [ConcurrentMap](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentMap.html).    
+**Example:**
+```java
+class AvoidNonAtomicIfModifyOnConcurrentMap {
+    void badPut(ConcurrentMap<String, String> accountMap, String accKey, String account) {
+        if (!accountMap.containsKey(accKey)) {
```
stokpop: Other Maps than ConcurrentMap used in a multi-threaded way can also be flagged? So if a class has any synchronized or imports from concurrent library? Or is this covered by other rule?

docs/JavaCodePerformance.md
```diff
@@ -2681,6 +2681,38 @@ class LombokInstanceLockUsedGood {
   private Map<String, String> cachedData = new HashMap<>();
 }
 ```
+#### TUTC15
+**Observation: Non-atomic if-modify is used on a ConcurrentMap.**
```
stokpop: What if it is protected by synchronized or explicit lock? Can cause false positives? Or do we still prefer the combined call even within a protected code block?

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidNonAtomicIfModifyOnConcurrentMap.xml
```diff
@@ -0,0 +1,76 @@
...
+        accountMap.computeIfAbsent(accKey, this::getAccount);
+    }
+
+    private String getAccount(String key) {
+        return "myAccount";
+    }
+
+    String bad3(ConcurrentMap<String, String> accountMap, String accKey, String account) {
+        if(accountMap.containsKey(accKey)) { // bad 3
+            return accountMap.get(accKey); // strictly not a modify, so actually not that bad
```
stokpop: ... but it could still return (unexpected) null in some cases I guess? So that is why it is mildly bad :-) ? Make more clear.

stokpop (changes requested):
see questions

docs/JavaCodePerformance.md
jborgers: I guess so. I'm not changing it, because the rule checks for containsKey.

docs/JavaCodePerformance.md
jborgers: Which ones do you mean? Since ConcurrentMap is an interface, it covers:
All Known Subinterfaces:
[ConcurrentNavigableMap](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentNavigableMap.html)<K,V>
All Known Implementing Classes:
[ConcurrentHashMap](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentHashMap.html), [ConcurrentSkipListMap](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ConcurrentSkipListMap.html)

docs/JavaCodePerformance.md
jborgers: Right. I think the proper way is to apply the atomically combined operation ConcurrentMap offers. 
And if within a guarded block, it is not a bug, but still the preferred way to me. If someone doesn't like it, then suppress.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/concurrent/xml/AvoidNonAtomicIfModifyOnConcurrentMap.xml
jborgers: Right, good point! Another thread can remove in between (the if and the get)

docs/JavaCodePerformance.md
stokpop: so a `get` call with `null` check can also be used as an "contains" check, not sure if you want to also check for that in this rule?

docs/JavaCodePerformance.md
stokpop: I meant regular java.util.Map, but there might be a rule already that says to use a ConcurrentMap when the class contains "multi-threading hints" such as `synchronized` keyword or uses of `Lock`.

docs/JavaCodePerformance.md
stokpop: ok

docs/JavaCodePerformance.md
jborgers: That is similar indeed. Added support for that case.

docs/JavaCodePerformance.md
jborgers: Yeah, that is for another rule

stokpop (approved):

PR #671, opened 2025-12-19
Rule request: configure Netty ConnectionProvider explicitly: NettyConnectionProviderWithoutMaxOrTimeout #552

### LINKED ISSUE #552: Rule request: configure Netty ConnectionProvider explicitly: NettyConnectionProviderWithoutMaxOrTimeout
If you don't specify the number of connections for the Connection Provider a default will be applied, which happens to be max(#CPU, 8) *2. This is usually not desired, often a higher value, at least made clear, explicitly defined in the code. A too low value will result in throttling,  delays and too long response times.  

```
import reactor.netty.resources.ConnectionProvider;

public ConnectionProvider connectionProvider() {
        return ConnectionProvider.builder("gupProvider")
                //.maxConnections(props.getMaxConnections()) // <-- this line was missing
                .pendingAcquireTimeout(Duration.ofMillis(props.getPendingAcquireTimeoutMs()))
                .pendingAcquireMaxCount(props.getPendingAcquireMaxCount())
                .maxIdleTime(props.getIdleTimeSeconds())
                .build();
  }
```
Investigate what methods should be present, be explicitly defined.

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -650,6 +650,34 @@ class ValidatingClientHttpRequestInterceptorGoodExample implements ClientHttpReq
...
 
+#### IBI26
+**Observation: The Netty HTTP connection provider uses the default maximum connections or the default pending acquire timeout.**   
+**Problem:** If a Reactor Netty Connection Provider is built without an explicit `maxConnections`, this defaults to `max(#CPU, 8) *2`.
+This is often too low, leading to throttling, delays and too long response times.   
+If the provider is built without an explicit `pendingAcquireTimeout`, this defaults to `45s`.
+This is often too high, leading to unnecessary long waiting times.   
+**Solution:** Define both `maxConnections` and `pendingAcquireTimeout` explicitly, with proper values, for instance, `50` and `300 [ms]` respectively.   
+**Notes:** 
+* Different from Apache HttpClient, Reactor Netty uses only one HttpClient with one Connection Provider per route.
```
stokpop: Not clear to me: you mean "each Reactor Netty HttpClient uses one Connection Provider *per route*" and a Apache HttpClient uses one Connection Provider for all routes? So each remote host in Reactor Netty will have its own connection pool (with max number of connections) and Apache shares one connection pool with all routes (with a "per route max" and "overall max" number of connections)?

docs/JavaCodePerformance.md
```diff
@@ -650,6 +650,34 @@ class ValidatingClientHttpRequestInterceptorGoodExample implements ClientHttpReq
...
+**Solution:** Define both `maxConnections` and `pendingAcquireTimeout` explicitly, with proper values, for instance, `50` and `300 [ms]` respectively.   
+**Notes:** 
+* Different from Apache HttpClient, Reactor Netty uses only one HttpClient with one Connection Provider per route.  
+* For calculating pool size, max connections, see: [IBI03](#IBI03)
+* For reasonable timeout values, see: [IBI10](#IBI10)   
+
+**Example**
+```java
+public ConnectionProvider connectionProviderBad() {
+    return ConnectionProvider.builder("gupProvider").build();  // bad
```
stokpop: What is gup?

docs/JavaCodePerformance.md
```diff
@@ -650,6 +650,34 @@ class ValidatingClientHttpRequestInterceptorGoodExample implements ClientHttpReq
 ```
 **See:** [ClientHttpRequestInterceptor javadoc](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/client/ClientHttpRequestInterceptor.html)  
 
+#### IBI26
+**Observation: The Netty HTTP connection provider uses the default maximum connections or the default pending acquire timeout.**   
+**Problem:** If a Reactor Netty Connection Provider is built without an explicit `maxConnections`, this defaults to `max(#CPU, 8) *2`.
```
stokpop: What was confusing to me: when you *just* use `HttpClient.create()` and *do not* supply an explicit `ConnectionProvider`, other default are used via a shared `ConnectionProvider` with max 500 connections (which are they call max active channels in the docs).

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/remoting/xml/NettyConnectionProviderWithoutMaxOrTimeout.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<test-data
+        xmlns="http://pmd.sourceforge.net/rule-tests"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://pmd.sourceforge.net/rule-tests http://pmd.sourceforge.net/rule-tests_1_0_0.xsd">
+    <test-code>
+        <description>violation: non-Closeable HttpClient</description>
```
stokpop: update description?

src/main/resources/category/java/remoting.xml
```diff
@@ -1567,4 +1567,41 @@ class ValidatingClientHttpRequestInterceptorGoodEx implements ClientHttpRequestI
         </example>
     </rule>
 
+    <rule name="NettyConnectionProviderWithoutMaxOrTimeout"
+          message="Netty ConnectionProvider has no defined maxConnections or pendingAcquireTimeout."
+          class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
+          language="java"
+          externalInfoUrl="${doc_root}/JavaCodePerformance.md#ibi26">
+        <description>Problem: If a Reactor Netty Connection Provider is built without an explicit `maxConnections`, this defaults to `max(#CPU, 8) *2`.
```
stokpop: add space to `* 2`?

docs/JavaCodePerformance.md
stokpop: add space to `* 2`?

stokpop (changes requested):
some small comments and suggested clarification in docs

src/main/resources/category/java/remoting.xml
```diff
@@ -1567,4 +1567,41 @@ class ValidatingClientHttpRequestInterceptorGoodEx implements ClientHttpRequestI
...
+            This is often too low, leading to throttling, delays and too long response times.
+            If the provider is built without an explicit `pendingAcquireTimeout`, this defaults to `45s`.
+            This is often too high, leading to unnecessary long waiting times.
+            Solution: Define both `maxConnections` and `pendingAcquireTimeout` with proper values, for instance, `50` and `250 [ms]` respectively.
```
stokpop: in docs example it was 300 ms, make same?

docs/JavaCodePerformance.md
jborgers: good point, clarified

docs/JavaCodePerformance.md
jborgers: arbitrary name, changed to 'myProvider'

docs/JavaCodePerformance.md
jborgers: Yes, indeed. That is another case, this rule is about the case that you do supply it. Maybe we should create another rule for that. Always use a connection provider explicitly?

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/remoting/xml/NettyConnectionProviderWithoutMaxOrTimeout.xml
jborgers: check

src/main/resources/category/java/remoting.xml
jborgers: check

docs/JavaCodePerformance.md
jborgers: check

src/main/resources/category/java/remoting.xml
jborgers: check

src/main/resources/category/java/remoting.xml
```diff
@@ -1567,4 +1567,41 @@ class ValidatingClientHttpRequestInterceptorGoodEx implements ClientHttpRequestI
...
+                <value><![CDATA[
+//TypeExpression[pmd-java:typeIs('reactor.netty.resources.ConnectionProvider')]
+[not(ancestor::MethodCall[@MethodName='maxConnections']) or not(ancestor::MethodCall[@MethodName='pendingAcquireTimeout'])]
+    ]]></value>
+            </property>
+        </properties>
+        <example>
+            <![CDATA[
+public ConnectionProvider connectionProviderBad() {
+    return ConnectionProvider.builder("gupProvider").build(); // bad
```
stokpop: still `gupProvider` in example (2x)

stokpop: (also in the unit tests, but those ore more hidden so you could leave that)

docs/JavaCodePerformance.md
stokpop: maybe yes; the 500 is in most cases "better" than the other default which is much lower: still good to make settings explicit instead of using defaults.

I thought to maybe add a little explanation in the docs about this to avoid confusion for now.

stokpop (changes requested):
small followup answers/change requests

src/main/resources/category/java/remoting.xml
jborgers: check

jborgers:
there is a link to the netty docs and I think it is out of scope now. and a little explanation may be half-baked and also trigger new questions. Better deal with it fully in a separate rule with full explanation.

stokpop (approved):

PR #691, opened 2026-02-06
Fix issue 651 - additional changes AvoidMultipleConcatStatements

stokpop:
* merge
* found hashCode false positive

* now exclude on hashCode fun, but that if obvious String concat is inside hashCode it is not found: added INACTIVE test for this, to be fixed when needed


jborgers:
Seems like a hack. How about this pattern in non-hashCode fun's?

jborgers (commented):
Is this only for hashCode? Other methods working with Int?

PR #698, opened 2026-03-29
pmd7-issue697 Extend AvoidLoadingAllFromFile with all major load‑all‑into‑memory patterns

### LINKED ISSUE #697: Fix Request: Extend AvoidLoadingAllFromFile with all major load‑all‑into‑memory patterns
Our current PMD rule `AvoidLoadingAllFromFile` only detects `Files.readAllBytes()` and `Files.readAllLines()`. These patterns are indeed dangerous for large file uploads, but the encountered codebase also used several other APIs and classes that cause the same memory‑explosion risk.

To fully enforce our streaming‑only approach and prevent future OOM issues, the rule needs to be extended to catch additional memory‑loading methods and classes.

**Problems observed in the codebase**
The following APIs and constructions are equally dangerous because they read entire files or streams into heap memory:

1. Java standard library
```
InputStream.readAllBytes()
Files.readString(...)
ByteArrayInputStream(byte[])
```

2. Apache Commons IO

`IOUtils.toByteArray(InputStream|URL|Path)`

3. In‑memory buffer classes

`ByteArrayOutputStream.toByteArray()`
`ResizableByteArrayOutputStream (Spring)` — used in the encountered implementation and can grow to multi‑MB size silently.

All of these patterns conflict with our requirement to keep large uploads streaming and avoid loading full content into memory.

**Request**
Extend the existing PMD rule so it also detects:

```
InputStream.readAllBytes()
Files.readString(...)
IOUtils.toByteArray(...)
any allocation of ByteArrayInputStream
any ByteArrayOutputStream.toByteArray() usage
any ResizableByteArrayOutputStream.toByteArray() usage
```

jborgers:
Extend AvoidLoadingAllFromFile with all major “load‑all‑into‑memory” patterns #697

### DIFF
--- a/src/main/resources/category/java/common.xml
+++ b/src/main/resources/category/java/common.xml
@@ -2139,13 +2139,11 @@ class Foo {
         </example>
     </rule>
 
-
     <rule name="AvoidLoadingAllFromFile"
-        language="java"
-        message="Files.readAll methods and similar APIs load all bytes into memory: a risk of memory problems."
-        class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
-        externalInfoUrl="https://github.com/jborgers/PMD-jPinpoint-rules/tree/pmd7/docs/JavaCodePerformance.md#isio03">
-
+          language="java"
+          message="Files.readAll methods and similar APIs load all bytes into memory: a risk of memory problems."
+          class="net.sourceforge.pmd.lang.rule.xpath.XPathRule"
+          externalInfoUrl="${doc_root}/JavaCodePerformance.md#isio03">
         <description>
             Problem: APIs like Files.readAllBytes, Files.readAllLines, Files.readString,
             InputStream.readAllBytes, IOUtils.toByteArray, ByteArrayInputStream,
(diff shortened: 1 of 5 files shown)


stokpop:
Looks good to me.

I have a worry: will this not produce to many falls positives for 1-off file loading and/or small files?

Will be difficult to filter, maybe allow in @Configuration like reading in security certs? 
Or flag reading of large files only in @Service @Controller code where it will be repeatably called? (Don't think this is doable: can call helper methods in other classes that do the file loading).

src/main/resources/category/java/common.xml
```diff
@@ -2141,47 +2141,59 @@ class Foo {
...
-            Often, functionality can be achieved in a streaming way.
-            Note: not a problem for small files.
+        <description>
+            Problem: APIs like Files.readAllBytes, Files.readAllLines, Files.readString,
+            InputStream.readAllBytes, IOUtils.toByteArray, ByteArrayInputStream,
+            ByteArrayOutputStream.toByteArray, and ResizableByteArrayOutputStream.toByteArray
+            load entire content into heap memory. This risks OutOfMemoryError, long GC pauses,
+            and slow responses when processing large files.
+            Solution: Use streaming APIs end-to-end, do not buffer whole files in memory.
+            Note: Not a problem for small files.
```
stokpop: Should we explicitly say that you can use @SuppressWarning in these cases with a proper comment to clarify.

E.g. 
@SuppressWarning("pmd.AvoidLoadingAllFromFile") // expected files size is small: most 1kb

stokpop (changes requested):
looks good, bit worried about too many false positives (see previous comment)

src/main/resources/category/java/common.xml
jborgers: Good point, will add to description.

stokpop (approved):

PR #706, opened 2026-05-13
Fix for #704: false positive on UnconditionalCreatedLogArguments

### LINKED ISSUE #704: Fix Request: false positive on UnconditionalCreatedLogArguments
The rule should not violate on use of size() on java.util implementation types because this is typically very cheap and can usually be inlined by the compiler.

`LOGGER.debug("Created number of fluffs: {}", fluffs.size());`

The same holds for `isEmpty()`, `Map.containsKey(k)`(except TreeMap), `Set.contains(e)` (except TreeSet)

jborgers:


stokpop (commented):
Looks good to me, might mention that side effects, or returning a value just for logging seems inappropriate.

stokpop (approved):

PR #710, opened 2026-08-07
Rule for Request #707: AvoidRecreatingMessageDigests - MessageDigest.getInstance is expensive because of classloading

### LINKED ISSUE #707: Rule Request: AvoidRecreatingMessageDigests  - MessageDigest.getInstance is expensive because of classloading
MessageDigest.getInstance is expensive because of:
- classloading
- provider lookup
- synchronization

Under load it can result in high contention.

This rule is similar to AvoidRecreatingSecurityProviders.

Since MassageDigest is not thread-safe, it cannot be shared among threads.

Since virtual threads, ThreadLocals are no longer a solution because virtual threads are short lived. ScopedValues are not a solution since they are immutable.

Using MessageDigest.clone() is an efficient solution to avoid the expensive part. A clone() is typically supported by the provider implementation, but not guaranteed.

Apache commons codec DigestUtils does not do such optimization and is therefore to avoid.

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -2259,7 +2259,48 @@ class Foo {
...
+**Note:** An addProvider call inside a static main method or inside a @PostConstruct annotated method is *not* reported as a violation since it is assumed to be called only once. 
+
+#### IUOSF02
+
+**Observation: A MessageDigest object is re-created, that is, created in a method called more than once.**   
+**Problem:** Creating a MessageDigest object is expensive because of provider lookup, loading of algorithms and other classes.
+Additionally, it uses synchronized which leads to lock contention when used with multiple threads, resulting in response time spikes.  
+**Solution:** Since MessageDigest is not thread-safe, an instance cannot simply be shared among threads. Create the MessageDigest only once and clone it for each use. 
+MessageDigest.clone() is efficient and typically supported by the provider, yet note that support is not guaranteed.   
+**Note:** Apache Commons codec DigestUtils does not do such optimization. Only methods which create a MessageDigest under the hood are, therefore, to avoid.
```
stokpop, line 2271: By "does not do such optimization" you mean the util methods do not make use of clone()? "Only methods ..." sentence is very clear... better: "Avoid methods that create a MessageDigest under the hood."

docs/JavaCodePerformance.md
```diff
@@ -2259,7 +2259,48 @@ class Foo {
...
+**Note:** Apache Commons codec DigestUtils does not do such optimization. Only methods which create a MessageDigest under the hood are, therefore, to avoid.   
+**Rule name:** AvoidRecreatingMessageDigests.    
+**Example:**
+```java
+import java.security.MessageDigest;
+import org.apache.commons.codec.digest.DigestUtils;
+import static org.apache.commons.codec.digest.MessageDigestAlgorithms.SHA_256;
+
+class Foo {
+    MessageDigest mdField = MessageDigest.getInstance("SHA-256");
```
stokpop, line 2280: Think of better name for `mdField`, maybe `messageDigestDonor`? It will be used as copy-paste example.

Some for `mdLocal` below... I think good names for tests but not for examples.

Can the messageDigestDonor also be static final? This will load the message digest at class loading... maybe not what you want.

rulesets/java/jpinpoint-rules.xml
```diff
@@ -888,6 +888,59 @@ class Good {
...
+        <priority>2</priority>
+        <properties>
+            <property name="tags" value="cpu,io,jpinpoint-rule,performance,sustainability-high" type="String" description="classification"/>
+            <property name="xpath">
+                <value><![CDATA[
+//MethodDeclaration[not((@Name='main' and @Static=true()) or ModifierList/Annotation/@SimpleName='PostConstruct')]//(
+    LocalVariableDeclaration[ClassType[pmd-java:typeIs('java.security.MessageDigest')]]/VariableDeclarator/MethodCall[@MethodName='getInstance']
+  | MethodCall[starts-with(@MethodName,'digest')]/ConstructorCall[pmd-java:typeIs('org.apache.commons.codec.digest.DigestUtils')]
+  | LocalVariableDeclaration[ClassType[pmd-java:typeIs('java.security.MessageDigest')]]//MethodCall/TypeExpression[pmd-java:typeIs('org.apache.commons.codec.digest.DigestUtils')]
+  | MethodCall[starts-with(@MethodName,'sha') or starts-with(@MethodName,'md')]/TypeExpression[pmd-java:typeIs('org.apache.commons.codec.digest.DigestUtils')]
```
stokpop, line 909: Document why these two method prefixes `sha` and `md` are ok. Not too brittle? There are also `shake` prefixes in DigestUtils... oh those also start with `sha`.

docs/JavaCodePerformance.md
```diff
@@ -2259,7 +2259,48 @@ class Foo {
...
-**Note:** An addProvider call inside a static main method or inside a @PostConstruct annotated method is not reported as a violation since it is assumed to be called only once. 
+**Note:** An addProvider call inside a static main method or inside a @PostConstruct annotated method is *not* reported as a violation since it is assumed to be called only once. 
+
+#### IUOSF02
+
+**Observation: A MessageDigest object is re-created, that is, created in a method called more than once.**   
+**Problem:** Creating a MessageDigest object is expensive because of provider lookup, loading of algorithms and other classes.
+Additionally, it uses synchronized which leads to lock contention when used with multiple threads, resulting in response time spikes.  
+**Solution:** Since MessageDigest is not thread-safe, an instance cannot simply be shared among threads. Create the MessageDigest only once and clone it for each use. 
+MessageDigest.clone() is efficient and typically supported by the provider, yet note that support is not guaranteed.
```
stokpop, line 2270: From the javadoc: "Implementations are free to implement the Cloneable interface. Client applications can test cloneability by attempting cloning and catching the CloneNotSupportedException"

stokpop: Question: does clone work for the BouncyCastle provider?

stokpop (changes requested):
Some questions and possible improvements.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidRecreatingMessageDigests.xml
```diff
@@ -0,0 +1,94 @@
...
+    byte[] bad1DigestUtils() {
+        return new DigestUtils(SHA_256).digest(dataToDigest); // bad2
+    }
+
+    String bad2DigestUtils() {
+        return new DigestUtils(SHA_256).digestAsHex(dataToDigest); //bad3
+    }
+
+    MessageDigest good1DigestUtils() {
+        return DigestUtils.getMd5Digest(); // okay, can be used to initialize once
```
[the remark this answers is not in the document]
jborgers: We don't know from where this method is called, could be to initialize once, like in a PostConstruct, so we cannot determine that it is bad. Better avoid false positives than missing czses.

stokpop (dismissed):
Approved, with two optional remarks to be fixed.

PR #711, opened 2026-08-25
Fix For Request #708: missing case in AvoidReStreamingEnumValues

### LINKED ISSUE #708: Fix Request: missing case in AvoidReStreamingEnumValues
This case is not found:
```
private Kind convertToKind(String type) {
		return Arrays.stream(Kind.values())
				.filter(k -> k.value().equals(type))
				.findFirst()
				.orElse(null);
}
```

jborgers:


docs/JavaCodePerformance.md
```diff
@@ -3115,9 +3115,11 @@ Set<YourEnumType> set = EnumSet.allOf(YourEnumType.class);
...
-**Problem:** the time to find element is O(n); n = the number of enum values. This identical processing is executed for every call. Considered problematic when n > 3.     
+**Problem:** the time to find an element is O(n); n = the number of enum values. This identical processing is executed for every call. Considered problematic when n > 3.     
 **Solution:** use a static field-to-enum-value Map. Access time is O(1), provided the [hashCode is well-defined](http://www.ibm.com/developerworks/library/j-jtp05273/).
-For one String field, usually toString returns that field. Consider to implement a fromString method to provide the reverse conversion by using the map, see the following examples:   
+For one String field, usually toString returns that field. Consider implementing a fromString method to provide the reverse conversion by using the map, see the following examples.   
+**Rule name:** AvoidReStreamingEnumValues   
+**Note:** The rule will flag for cases with more than 3 enum values, in case it occurs in a method inside the enum itself. Outside the enum it does not consider the number of enum values.
```
stokpop: I assume it does not consider the number of enums because it cannot count the number of values of the enum in this case? Better add the method to the original enum anyway (if possible).

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidReStreamingEnumValues.xml
```diff
@@ -28,6 +28,12 @@ public enum BadFruit {
         return Arrays.stream(BadFruit.values()).filter(f -> f.toString().equals(name)).findFirst(); // bad: iterates for every call, O(n) access time
     }
 }
+
+class BadFruitHelper {
+    BadFruit toBadFruit(String name) {
+        return Arrays.stream(BadFruit.values()).filter(v -> v.toString().equals(name)).findFirst().orElse(null); // bad: iterates for every call
```
stokpop: Does the rule not flag this method when the BadFruit is *not* an enum? Maybe add false positive regression test?

stokpop (changes requested):
Two remarks to check.

docs/JavaCodePerformance.md
jborgers: Indeed. Agree that it seems better to add to the enum, however, there may be reasons not to do so, maybe it is style. At least not a performance issue.

src/test/resources/com/jpinpoint/perf/lang/java/ruleset/common/xml/AvoidReStreamingEnumValues.xml
jborgers: Just when BadFruit is an enum, then `.values()` is predefined and we know it is O(n).
Not sure what you mean with a false positive regression test? A test of BadFruitClass not being flagged? Not sure what that adds.

stokpop: To prevent `NonEnumClass.values()` to be matched I guess... via `pmd-java:matchesSig('_#values()')`

jborgers: Good point, fixed

stokpop (approved):
LGTM

